PR #3, opened 2026-07-18
feat(openai): add chat/completions tool-calling support

chaudhryfaisal:
Bridge OpenAI /v1/chat/completions requests into the existing think/tool-aware pipeline (chatml_prompt + tools_preamble + ToolConstrainer + StreamSplitter + parse_tool_call). Adds openai_tools_json and openai_msgs in api_common.h to map the OpenAI wire shape (function.arguments as a JSON string, role:"tool", role:"developer") onto the internal Msg/tools format, tolerating malformed single tools instead of failing the whole request. The routed_chat path is gated strictly on a "messages" body so /v1/completions and raw-"prompt" chat keep byte-for-byte text-only behavior. Includes CPU-only integration + unit test harnesses with a drift check to keep the extracted server logic in sync.

### DIFF
--- a/src/api_common.h
+++ b/src/api_common.h
@@ -311,6 +311,120 @@ inline std::vector<Msg> anthropic_msgs(const json& body) {
     return msgs;
 }
 
+// OpenAI chat/completions tools -> the same {"type":"function","function":{...}}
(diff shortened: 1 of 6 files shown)


signalnine:
Looks great, thanks!

PR #5, opened 2026-07-20
feat(auth): add opt-in API key authentication

chaudhryfaisal:
Add an optional shared-secret gate for the server. Keys can be supplied via --api-key (repeatable), --api-key-file (one key per line, # comments ignored), or the Q27_API_KEY env var; any configured key is accepted. Both the OpenAI Bearer and Anthropic x-api-key header conventions are supported, with constant-time comparison and per-family 401 error shapes. /health stays unauthenticated for infra health checks. Binding non-loopback with no key configured prints a warning but is not refused, preserving reverse-proxy deployments that terminate auth themselves. Includes CPU unit tests and real httplib integration tests for the auth helpers and pre-routing handler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,10 +113,12 @@ make
 
 # 4. serve -- zero config; defaults resolve the full measured stack
 #    and --ctx auto-sizes to your VRAM (see Serving for escapes).
-#    Binds 127.0.0.1 only: the server has NO auth. To reach it from
-#    other machines or from containers (Claude Code in docker resolves
-#    the host via the bridge, not loopback), opt in explicitly:
-#      --host 0.0.0.0
+#    Binds 127.0.0.1 only. To reach it from other machines or from
+#    containers (Claude Code in docker resolves the host via the
+#    bridge, not loopback), opt in explicitly -- and set an API key
+#    at the same time, since --host 0.0.0.0 with no key accepts
+#    unauthenticated requests from anyone who can reach the port:
+#      --host 0.0.0.0 --api-key <your-key>
 ./build/q27-server ../models/qwen36-27b-mtp/qwen36-27b-mtp.q27 \
   ../models/qwen36-27b-mtp/qwen36-27b-mtp.tok --port 8080
 ```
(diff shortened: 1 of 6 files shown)


signalnine:
Nice PR -- opt-in, pre-routing so nothing bypasses it, keys never logged, and the SECURITY-MODEL.md additions are honest about what it isn't (not TLS / rate-limiting / multi-tenant). `secure_compare` is properly constant-time, and both suites pass here (unit + the httplib integration). One thing to fix before I merge:

`api_key_valid` uses:

```cpp
for (auto& k : keys) any = any || secure_compare(provided, k);
```

The `||` short-circuits -- once a key matches, the remaining `secure_compare` calls are skipped, which undercuts the "always scans every key ... time depends only on key count/length" note right above it. It's minor (it only leaks *which* configured key matched, via timing, and only to someone who already holds a valid key -- a wrong guess never sets `any`, so it always scans the whole set; not a key-guessing vector), but it contradicts the stated invariant. One-char fix:

```cpp
for (auto& k : keys) any |= secure_compare(provided, k);
```

`|=` is bitwise, so it doesn't short-circuit -- every key is compared regardless of which matched.

Optional nit (not blocking): `test_auth_integration.cpp` keeps a verbatim copy of the `server.cu` handler. Worth a comment that they must stay in sync -- a real bypass bug in `server.cu` would still pass the test if only the copy were correct.

Heads up on rebase: this branched before v0.3.5, which also touched `api_common.h` (the tool-drift fixes are elsewhere in the file, so it should merge clean, but pull master in). Flip the `||` and I'll merge.

signalnine:
Merged, thanks.

Verified on the merge: the constant-time scan is `any |= secure_compare(...)` now (walks every configured key regardless of match), which is what the "no early return" comment promised. Clean disjoint merge against master, `server.cu` builds green, and all four CPU suites pass -- `test_auth`, `test_auth_integration` (real loopback socket), plus `test_tool_drift` / `test_stream_split` as a regression check since we both touched `api_common.h`.

Dual `x-api-key` + `Bearer` support lands exactly where we need it -- Claude Code sends one, Kilocode/OpenAI-compatible clients the other, so neither family needs special config. And the SECURITY-MODEL addendum is honest about what this is and isn't (shared secret, not TLS/rate-limiting/tenancy) rather than overselling it.

One follow-up, non-blocking: the integration test keeps a verbatim copy of the pre-routing handler instead of exercising server.cu's, so a bug introduced only in server.cu's copy wouldn't trip it (the NOTE comment on that line also got mangled in an edit -- worth a cleanup pass). Fine as-is for now.

PR #9, opened 2026-08-04
fix: commit reasoning budget transitions through engine state

### LINKED ISSUE #7: reasoning budget closes the output splitter but not decoder state
Current master enforces the reasoning budget by feeding </think> to StreamSplitter, but the close tokens are never committed through the decoder or KV state.

After the budget expires, the output parser switches to the answer channel, but generation continues from the previous reasoning trajectory. In a reproduction with a budget of 1, usage correctly reported reasoning_budget_exceeded: true, while the visible answer began with continued reasoning rather than the requested answer.

I have a focused CUDA-only fix that:
 - commits the tokenizer’s exact close sequence through decoder state;
 - closes a zero budget before generating a reasoning token;
 - safely handles the transition when speculative decoding is enabled; and
 - adds focused regressions for the resulting output and usage.

I'll open a draft PR, but a decision point: would you prefer the forced close to be represented as a small engine operation, or handled at the decode-task/conductor boundary?

manthedan:
Fixes #7.

## Problem

The server enforced a reasoning budget by feeding `</think>` only to `StreamSplitter`. That changed the public channel without committing the close through decoder/KV state, so generation continued from the unclosed reasoning trajectory and the next visible answer token could still be reasoning.

## Change

- Make one `Engine` operation own forced close-token commitment, pending-token replacement, speculative-lane invalidation, and identical solo/conductor behavior.
- Check positive greedy budgets after the accepted speculative round, allowing bounded round-width overshoot rather than rewinding for exact cap precision.
- Close a per-request zero budget before the first model-generated reasoning token; preserve CLI `--think-budget 0` as the unbounded opt-out.
- Use the existing refinish path only when retaining the whole round would consume reserved close/answer capacity.
- Select any capacity-limited retained prefix before the stateful tool scanner runs, then commit the same prefix to both observers.
- Keep nested Responses status, `incomplete_details`, `response.incomplete`, and reasoning usage truthful in streaming and non-streaming paths.

## Product decisions

The default scope is intentionally prompt-seeded reasoning. Standard thinking-enabled serving starts inside the template-provided reasoning block and remains covered. A later model-generated block is capped only when explicitly armed through a positive CLI budget or request-level budget.

Bounded sampled requests intentionally use one-token acceptance rounds so every accepted token is a coherent budget-observation point. This disables speculative acceptance for the bounded request. The maintainer's reviewed RTX 5090 measurement was 56.9 t/s bounded versus 143.1 t/s unbounded, a 2.51x reduction. Unbounded sampled requests retain normal speculative width. Restoring sampled speculative width without weakening transition semantics is separate follow-up work.

Greedy checks occur after an accepted speculative round, so the reasoning count may exceed the configured limit by that round's accepted width. A natural close later in that round ends normally rather than reporting a budget trip.

## Verification

Current-upstream stabilized tip: `e7fde25a2e6f92239b67e187f804c34a1b6d6b62`, rebased onto upstream `5ce983d059b7ba78dfef365a9b141d313ae33b46`. Its `src/` tree is byte-identical to the previously reviewed `1de550a7bf81`; the upstream refresh adds the corrected per-architecture sampling gate and BUILDLOG records.

Host gates on the stabilized tip:

- conductor forced-member and queue-provenance behavior
- think resolution, admission, and enforcement
- stream splitting
- OpenAI bridge behavior
- chat-handler integration and stream/non-stream parity
- `git diff --check`

Ampere / `sm_86`, CUDA 12.8.93, our local RTX 3090:

- built `q27`, `q27-server-w8`, `test_kernels`, and `fused_smoke`
- corrected `CANON_ARCH=sm86` sampling gate: canonical `6894254e3b1a184ee3802771ddd59c2b` EXACT; all five sampling gates PASS
- full model-backed `test_kernels`: `ALL PASS`
- live width-8 `Q27_BATCH=1` sampled Responses request: budget 4 reported exactly four reasoning tokens, `reasoning_budget_exceeded: true`, and continued into answer text
- the full two-engine fused binary cannot instantiate two full-model engines on the 24 GiB card, including the width-8/greedy-only profile; this is the documented capacity boundary, not a candidate failure

Published exact-source fused evidence remains applicable because `src/` is unchanged by the refresh:

- dedicated A40: full model-backed kernels, fused/conductor identity, injected-error isolation, graph-cache identity, and live Responses matrix PASS
- maintainer RTX 5090: canonical `a2982c5197c627551b27d76a0a94b220` EXACT, sampling gates PASS, 352/352 model-backed kernels PASS, `ninv_test` PASS, and every `fused_smoke` leg byte-identical

## Independent concurrency pass

The new `callback_forced` write/read is synchronous on the conductor thread and is copied into the queue during that call. Cancellation is atomic. SSE/client disconnect continues draining until queue closure, so the handler frame cannot unwind while the conductor still owns the member. Accepted-member normal completion, cancellation, shutdown, and caught solo/fused host exceptions finish decode before the final queue close/fail edge. No callback can fire after that edge.

The pass also noted a pre-existing, production-unreachable registration-refusal bookkeeping path that skips `finish_decode`; the queue sink is never installed there, so the new `DecodeTask` capture is not exposed. That cleanup is outside this correctness fix.

## Follow-up

Add a sampled round-boundary observer that preserves coherent reasoning-budget transitions without pinning speculative acceptance to one token for the full bounded request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,10 +203,11 @@ Two serving-side additions since: **P16 persistent prefix cache** (opt-in,
 P8 stable prefix to disk -- restart TTFT **8.15 s -> 1.20 s** on a 26,700-token
 prompt, bitwise-identical continuations, with an optional pinned host-RAM tier
 above it (`--prefix-cache-ram-gb`). And a **reasoning-token budget**, ON by
-default whenever a `<think>` block is served: the block is force-closed at half
-the request's `max_tokens` and the model still answers, because an A/B over 240
-scenarios found unbudgeted block mode truncating 28 times against inline's 0
-while scoring level on raw accuracy. See Serving for both.
+default when the rendered prompt begins inside a `<think>` block: the block is
+force-closed at half the request's `max_tokens` and the model still answers,
+because an A/B over 240 scenarios found unbudgeted block mode truncating 28
+times against inline's 0 while scoring level on raw accuracy. Later model-
+generated reasoning is capped only when explicitly armed. See Serving for both.
 
 ### Reference numbers (v0.2.0, 2026-07-16, vanilla model, 5090)
 
@@ -609,17 +610,32 @@ answer. When a client omits `max_tokens` the server defaults it to 8192
 thinking trace wants more, set it explicitly.
 
 **Reasoning budget (2026-07-30).** Tokens generated inside a prompt-seeded
-`<think>` block are capped; on trip the block is force-closed and the model
+`<think>` block are budgeted; on trip the block is force-closed and the model
 answers with what it has. It is not a hard stop -- truncating the request
-instead would reproduce the failure the cap exists to prevent. The default is
+instead would reproduce the failure the budget exists to prevent. The default is
 **half the request's `max_tokens`**, so the answer still fits after the close.
-Ordinary no-think requests remain unbounded and retain full sampled speculative
-decode width. `--think-budget N` with `N>0` explicitly arms that cap even for a
-later model-generated block; `--think-budget 0` opts out. Requests can likewise
-arm or override the budget, gated behind `--request-think`:
+The prompt-seeded scope is intentional: standard thinking-enabled serving
+starts inside the template-provided reasoning block, while a later model-
+generated block is capped only when explicitly armed. `--think-budget N` with
+`N>0` arms that later-block cap; `--think-budget 0` opts out. Requests can
+likewise arm or override the budget, gated behind `--request-think`:
 `thinking.budget_tokens` (Anthropic), `thinking_token_budget` (OpenAI/Qwen), or
-`chat_template_kwargs.thinking_budget` (llama.cpp). A trip is reported as
-`usage.reasoning_tokens` and `usage.reasoning_budget_exceeded`.
+`chat_template_kwargs.thinking_budget` (llama.cpp). Chat Completions and
+Messages report a trip as `usage.reasoning_tokens` and
+`usage.reasoning_budget_exceeded`; Responses reports the same fields under
+`usage.output_tokens_details`.
+
+Greedy checks land after the accepted speculative round, so the reasoning count
+may exceed the configured limit by that round's accepted width. A natural close
+later in the same round ends normally rather than reporting a budget trip.
+
+Bounded sampled requests intentionally use one-token acceptance rounds so every
+accepted token is a coherent budget-observation point. This disables
+speculative acceptance for the bounded request: on the reviewed RTX 5090 run,
+the exact tip measured 56.9 t/s bounded versus 143.1 t/s unbounded, a 2.51x
+reduction. Unbounded sampled requests retain normal speculative width. Restoring
+speculative width without weakening transition semantics is follow-up work.
+
 
 Why it defaults on: an 11-pack / 240-scenario A/B (BUILDLOG 2026-07-28, rescored
 07-30) found unbudgeted block mode truncating **28 times against inline's 0**
--- a/README.md
+++ b/README.md
@@ -608,18 +608,18 @@ answer. When a client omits `max_tokens` the server defaults it to 8192
 (unified across all three API shapes, clamped to the context window); a long
 thinking trace wants more, set it explicitly.
 
-**Reasoning budget (2026-07-30, ON by default whenever a block is served).**
-Tokens generated inside the `<think>` block are capped; on trip the block is
-force-closed and the model answers with what it has. It is not a hard stop --
-truncating the request instead would reproduce the failure the cap exists to
-prevent. Default is **half the request's `max_tokens`**, so the answer still
-fits after the close; `--think-budget N` sets an absolute cap and
-`--think-budget 0` opts out. Requests can bound it too, symmetric with the
-enable conventions above and gated behind the same `--request-think`:
+**Reasoning budget (2026-07-30).** Tokens generated inside a prompt-seeded
+`<think>` block are capped; on trip the block is force-closed and the model
+answers with what it has. It is not a hard stop -- truncating the request
+instead would reproduce the failure the cap exists to prevent. The default is
+**half the request's `max_tokens`**, so the answer still fits after the close.
+Ordinary no-think requests remain unbounded and retain full sampled speculative
+decode width. `--think-budget N` with `N>0` explicitly arms that cap even for a
+later model-generated block; `--think-budget 0` opts out. Requests can likewise
+arm or override the budget, gated behind `--request-think`:
 `thinking.budget_tokens` (Anthropic), `thinking_token_budget` (OpenAI/Qwen), or
 `chat_template_kwargs.thinking_budget` (llama.cpp). A trip is reported as
-`usage.reasoning_tokens` and `usage.reasoning_budget_exceeded` -- a budget that
-fired silently would be no better than the inert `budget_tokens` this replaced.
+`usage.reasoning_tokens` and `usage.reasoning_budget_exceeded`.
 
 Why it defaults on: an 11-pack / 240-scenario A/B (BUILDLOG 2026-07-28, rescored
 07-30) found unbudgeted block mode truncating **28 times against inline's 0**
--- a/docs/BUILDLOG.md
+++ b/docs/BUILDLOG.md
@@ -9999,9 +9999,48 @@ malformed-tool fallback; both were fixed and regression-pinned. The final Codex
 run returned no accepted/actionable findings (`patch is correct`, confidence
 0.91).
 
-CUDA gate: CUDA 12.8.93 on Yukon successfully rebuilt the corrected source with
+CUDA gate: CUDA 12.8.93 on our local RTX 3090 successfully rebuilt the corrected
+source with
 `make -B NVCC=/home/ddbb/.local/cuda-12.8/bin/nvcc build/q27-server-w8`.
-The compile emits only warnings in unchanged `prefill.cu` and `tokenizer.cpp`.
-No live-model result is claimed for this correction: the 3090 remained occupied
-by PID 2062415 (`evidence-wt/.venv/bin/python`, 21884 MiB), so taking the card
-would have disrupted another workload.
+The compile emitted only warnings in unchanged `prefill.cu` and `tokenizer.cpp`.
+No live-model result was claimed for this correction at the time because that
+GPU was occupied by another workload.
+
+## 2026-08-03 -- reasoning PR current-upstream stabilization and sampled pricing
+
+The five-commit reasoning candidate was rebased from `e2f150ca` onto current
+upstream `5ce983d0`. The `src/` tree is byte-identical to reviewed tip
+`1de550a7bf81`; the upstream change adds the corrected per-architecture sampling
+gate and BUILDLOG records. The PR documentation now states the intended default
+scope and sampled-budget price explicitly.
+
+Exact host gates on the rebased tree pass: conductor, think resolution and
+enforcement, stream splitting, OpenAI bridge, and chat integration. An
+independent concurrency pass found the changed queue-sink paths safe:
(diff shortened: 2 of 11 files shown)


signalnine:
Read the whole diff. The architecture is right and you improved on what I asked for in two places. Before it comes out of draft I want to flag two product decisions embedded in it, because neither is a bug fix and I don't think either has been priced yet.

## The mechanism is correct

Replacing the pending token at a round boundary and capping acceptance to one is a sound way to invalidate speculation: drafts produced under the old pending cannot commit, `h_next` is still the hidden after the last accepted token so drafting from (h_next, forced_id) is exactly "what follows the close", and the ctx-guard in `pre_round` runs before `stage_forced_pending`, so the `d_gen + t.Ph` write stays inside `max_ctx`. `generate()` keeps its old behaviour through the defaulted `external_task`, and `src/engine.cu` is untouched, so the CLI canonical path should be structurally inert.

Three things I want to name because they are better than what I proposed:

- `upload_accept_cap()` composing the tool cap and the reasoning cap into one upload fixes a latent clobber that was already there: `set_tool_constraint(-1)` wrote `h_cap0` straight to `d_accept_cap` and would have cleared a live reasoning cap. Composing them is the fix, and testing it against the stateful tool scanner is the right paranoia.
- `needs_solo_round` is the hazard I would most likely have missed. Fused drafting overwriting a staged forced id would have been a rare, batch-only corruption.
- Keeping `refinish` for the capacity boundary only, where retaining the round leaves no room for the close and an answer, is a better line than my "drop the rewind entirely". A capacity guard is a real constraint. Exactness was a preference, and I was wrong to treat the two as the same thing.

## Decision 1: bounded sampled requests lose speculative decode

`server.cu:756` sets `sampled_budget_watch` whenever a sampled request has a budget, and `engine.cuh:633` turns that into `set_reasoning_accept_cap(...)` for the life of the request. That pins `d_accept_cap` to 1, so a bounded sampled request accepts one token per round for its whole duration, not just across the transition.

I understand why: sampled rounds have no `on_round`, so capping acceptance is what makes every token a coherent boundary for the observer. It is the honest way to get correctness on that path. But MTP is most of why this engine is fast, and one-token rounds give up essentially all of it. Before this ships I'd like a number: bounded-sampled decode t/s against the same request unbounded, on any box. If it is the 2x-3x I expect, that belongs in the README next to the feature rather than being discovered later.

The follow-up worth considering, not in this PR: give the sampled path an `on_round`-equivalent so the observer gets its boundary without the cap.

## Decision 2: the default coverage narrowed

The README goes from "ON by default whenever a block is served" to capping a **prompt-seeded** block, with `--think-budget N>0` now required to arm the cap for a later model-generated block. That reads like a direct consequence of decision 1: don't arm by default, and most requests never pay the acceptance cap.

That trade is defensible, but it is a change to what the feature protects, and the 07-28 result is what the feature exists for. The serving path enables thinking through the chat template, so a served block is prompt-seeded and still covered, which is why I think this is fine. Say so explicitly in the PR if you agree, because right now the narrowing is visible only as a README diff and someone will later read it as documentation drift rather than an intended scope change.

## What I want before merge

- **A second review pass on the concurrency**, per the standing rule for state-heavy changes in this repo. This exact area has produced two UAF findings already, both on `TokenQueue` and `DecodeTask` lifetime across the conductor boundary, and the rule after the second one was that `fail()`/`close()` must be the last access to both `q` and `t`. Your sink now captures `t` as well as `q` and reads `t->callback_forced` per token. I traced it and believe it is safe, since that read is on the conductor thread synchronously inside `post_round`, the same thread and call that writes the flag. I would still rather someone else confirm it than have me be the only reader, given the history.
- **The numbers for decision 1.**
- The Blackwell canonical is mine to run. I'll build your tip on the 5090 and report, which closes the leg you have no hardware for.

On the canonical you were chasing: you already had it. See my comment on #7. `6894254e...` is the documented sm_86 value, your A40 reproduced it, and the mislabel was ours. I've since taught `tools/sampling_gate.sh` about the per-arch fork so the next contributor doesn't lose a night to it.

signalnine:
Blackwell leg done, on your exact tip `1de550a7bf81`. Clean worktree off the PR branch, nvcc 13, sm_120 RTX 5090, GPU otherwise idle.

## Gates

| gate | result |
|---|---|
| greedy canonical md5 | **`a2982c5197c627551b27d76a0a94b220` EXACT** |
| sampling gates 2-5 | ALL PASS |
| `test_kernels` (model-backed) | 352 PASS, 0 FAIL |
| `ninv_test` | ALL PASS |
| `fused_smoke` | solo/fused identity, conductor identity, A2 injected-error, graph passes 1 and 2 all byte-identical |

The canonical is the one you could not reach, and it is exact, so the greedy CLI path is bitwise-inert under this change on Blackwell as well as Ampere. The acceptance-vs-temp curve also came back identical to master's (greedy 2.58, T=0.3 2.88, T=0.7 2.86, T=1.0 2.75, T=1.5 1.00), which is a second, softer check on the sampled path not having drifted.

Between your A40 run and this one the PR now has both published architectures covered.

## The bounded-sampled number I asked for

I have the hardware, so rather than make you find a box, I measured it. Your tip, same server, `--request-think --think-budget 0`, one prompt, `temperature 0.7 top_p 0.95`, `max_tokens 700`, thinking enabled, n=2 per arm. The only difference between arms is whether the request carries `thinking.budget_tokens`.

| arm | tps | rounds | tok/round |
|---|--:|--:|--:|
| unbounded sampled | 143.1 | 284 | 2.46 |
| bounded sampled | 56.9 | 702 | 1.00 |

**2.51x slower, and the cap holds for the entire request**, not just across the transition. 702 rounds for 700 tokens is the acceptance cap doing exactly what the code says it does. Both reps in each arm agreed (56.9 / 56.9), so this is deterministic rather than noise.

The budget itself works correctly in that arm: `reasoning_tokens: 64`, `reasoning_budget_exceeded: true`, and the model still produced a full 925-character answer after the close. The feature does what it claims. It just costs 2.5x on sampled requests that arm it.

That is the number for the README. I don't think it blocks the merge, since narrowing the default is what keeps most traffic off this path, but a user who arms a per-request budget on a sampled request should be able to find out beforehand that they are turning off speculative decode.

Still worth a look as a follow-up, separate from this PR: giving the sampled path an `on_round` equivalent would get the observer its boundary without the cap, and would take this from 2.5x to roughly free.

## Where that leaves the PR

From my side the remaining items are the two in my earlier comment, and neither is mine to close:

1. Say explicitly in the PR that the default-coverage narrowing (README, prompt-seeded blocks) is intended scope rather than doc drift.
2. A second pair of eyes on the concurrency, per the repo's standing rule for state-heavy changes. My read is that the new `t->callback_forced` in the sink is safe, same thread and same call as the write, but this area has produced two UAF findings already and I would rather not be its only reader.

Everything else I can verify is green.

manthedan:
Thanks for the architecture correction, the Blackwell run, and the bounded-sampling measurement. That closes both canonical legs and gives us the number needed to make the sampled-budget tradeoff explicit.

I agree with both product decisions:

1. **Bounded sampled requests intentionally use one-token acceptance rounds in this PR.** Correct, coherent budget observation takes priority over speculative throughput when a sampled reasoning budget is active. I’ll document the measured result — 56.9 t/s bounded versus 143.1 t/s unbounded, a 2.51x reduction — next to the feature in README. Unbounded sampled requests retain normal speculative width.

2. **The default-coverage narrowing is intentional.** The default applies to prompt-seeded reasoning blocks. Standard serving enables thinking through the chat template, so its normal served reasoning block remains covered. A later model-generated reasoning block is unbounded unless explicitly armed through `--think-budget N` or a request-level budget. That keeps ordinary unarmed sampled requests off the one-token path rather than imposing the bounded-budget cost globally.

I’ll leave a sampled `on_round`-equivalent for a separate follow-up. Its goal will be to preserve coherent budget transitions while restoring speculative acceptance; I won’t add that redesign to this correctness PR.

I’m also completing the requested independent concurrency pass. The invariant under review is that every conductor exit runs `finish_decode` before queue closure or failure, and that queue closure is the synchronization edge after which the request thread may destroy both its queue and frame-local `DecodeTask`. The SSE/client-disconnect path must continue draining until that edge. I’ll report that review separately rather than claiming it complete here.

Finally, the PR now conflicts with current `master`, so before taking it out of draft I’ll rebase, retain the corrected per-architecture sampling gate, rerun the affected exact-tip host and Ampere gates, and refresh the PR body with the completed Ampere and Blackwell evidence.

Thanks again for running the Blackwell leg and for catching the performance consequence directly.

manthedan:
Independent concurrency pass complete on the current-upstream rebased tree.

The changed lifetime paths are sound:

- `callback_forced` is written immediately around `on_token` and read by the installed queue sink synchronously on the conductor thread; it is copied into the queue while that call is active, so there is no cross-thread data race on the flag.
- Request-thread cancellation uses the atomic `DecodeTask::cancel` flag.
- SSE/client disconnect sets cancellation but continues draining until the queue closes, so the handler frame and its frame-local `DecodeTask` cannot unwind while the conductor still owns the member.
- Accepted-member normal completion, cancellation, conductor shutdown, and caught solo/fused host exceptions run `finish_decode` before the final queue `close`/`fail` edge. No callback can run after that edge.
- `fail_member` nulls `mm.q` before `fail()`, and `fail()` remains the final access to both the queue and task state on the exception path.

The review also identified a pre-existing registration-refusal bookkeeping exception: when `register_member` races with `request_stop`, it fails the queue without `finish_decode`. The sink is deliberately not installed on that path, so the new `t->callback_forced` capture is never exposed and this is not a UAF or regression introduced by the PR. The production server's listener/conductor shutdown ordering makes the path unreachable there. I am keeping that separate from this correctness fix rather than broadening the PR.

Result for the changed concurrency/lifetime surface: no actionable defect found.

signalnine:
Both items closed from my side, and the concurrency pass is the shape I wanted: a stated invariant, each exit traced against it, and the one thing you found scoped out rather than folded in. Failing the queue without `finish_decode` when `register_member` races `request_stop` is worth its own issue whenever you feel like writing it, and you're right that the sink is never installed there so this PR doesn't touch it.

## My Blackwell evidence survives the rebase, and here is the proof

I gated `1de550a7bf81`, which the rebase replaced with `e7fde25a2e6f`. Rather than assume the result carried over or burn another twenty minutes of GPU on it, I diffed the build inputs:

```
git diff --stat 1de550a7b e7fde25a2
  README.md              |  40 ++++---
  docs/BUILDLOG.md       | 190 +++++++++++++++++++++++++++++++--
  tools/sampling_gate.sh |  29 ++++-
```

Every file under `src/` and the `Makefile` are byte-identical across the two tips. I checked each one by object id rather than eyeballing the stat. The only non-doc change is `tools/sampling_gate.sh`, which is my own per-arch commit that your rebase picked up.

So the binary built from `e7fde25a2` is the binary I gated, and the canonical, `test_kernels`, `ninv_test` and `fused_smoke` results stand for the current tip unchanged. Identity is a stronger argument than a second run would be: a re-run can pass for the wrong reason, whereas identical inputs cannot produce a different md5. Say the word if you want it re-run anyway for the record and I'll do it, but I don't think it buys anything.

## On the README

The 2.51x landed accurately, and I like that you kept the mechanism next to it rather than only the number. Two things there are better than what I asked for:

The note that greedy checks land after the accepted round, so the reasoning count can exceed the limit by that round's width, is the honest statement of the bounded-overshoot design. I argued for that overshoot without saying it would be user-visible in `usage.reasoning_tokens`. Documenting it is the right call.

Same for a natural close later in the same round ending normally rather than reporting a trip. That distinction only exists because the close is real decoder state now, and it would have been easy to leave it to be discovered.

Nothing blocking from me. The sampled `on_round` equivalent is the right follow-up and the right thing to keep out of a correctness PR.

PR #10, opened 2026-08-04
perf: restore sampled reasoning budget width

manthedan:
## Relationship to #9

Focused single-commit follow-up to merged PR #9. Current review commit: `6ad7ee10`.

## Problem

#9 intentionally gives every bounded sampled request one-token acceptance rounds so reasoning-budget transitions are coherent. The maintainer measured the cost on RTX 5090: 56.9 t/s bounded versus 143.1 t/s unbounded (2.51x slower).

## Change

- Run the request-local pre-emission `Engine::on_round` observer for sampled as well as greedy accepted rounds.
- Keep ordinary sampled speculative acceptance at its natural width.
- When a crossing round would consume reserved close/answer capacity, retain only the prefix through the budget trip and restore its hidden/KV/position state with the existing `refinish_round` path.
- Keep forced-transition rounds capped to one token so no speculative lane derived from the superseded pending token can commit.
- Treat a capacity-starved close as terminal cancellation: the retained sampled prefix remains coherent and the cancellation boundary prevents the provisional refinish argmax from driving another model decision.
- Preserve a natural close plus public answer/tool bytes that already fit in the same exact-cap round, including marker-buffered and incomplete-UTF-8 tails.
- Delete the obsolete full-request sampled acceptance cap and per-token sampled budget observer.

The budget remains round-granular: a naturally accepted round may overshoot the configured reasoning count, and a natural close later in that same round still wins. This PR changes throughput, not those semantics.

## Verification

Fresh host gates on the final commit:

- extracted `server.cu` / integration-fake parity: PASS
- `test_chat_completions_integration`: PASS, including solo/conductor sampled-boundary, exact-cap natural-close, marker/UTF-8 buffered-tail, EOS, stale-suffix, forced-conditioning, multi-token-width, and capacity-starved cancellation regressions
- `test_think_resolve`: PASS
- `test_depthctl`: PASS
- `test_conductor`: PASS

RTX 3090, CUDA 12.8.93, `sm_86`, rebuilt from `6ad7ee10`:

- `q27`, `q27-server-w8`, `test_kernels`, and the fused-smoke binary: BUILD PASS
- canonical greedy md5: `6894254e3b1a184ee3802771ddd59c2b` EXACT
- sampling gates 1-5: ALL PASS
- freshly rebuilt model-backed `test_kernels`: ALL PASS
- live `Q27_BATCH=1` and `Q27_BATCH=0` budget-1/max-2 sampled requests expose one reasoning token plus one answer token
- concurrent conductor requests remain isolated; zero budget exposes no reasoning; streaming and non-streaming report coherent public bytes and terminal usage

The two-full-engine `fused_smoke` runtime still exceeds the 24 GiB RTX 3090 capacity; this is recorded as a capacity boundary, not a runtime pass.

## Performance

The maintainer will run the final tip against #9 on RTX 5090 using the same prompt, sampling parameters, and protocol, so the README carries one matched before/after comparison. The earlier local 256-token run predates the final boundary hardening and is retained only as diagnostic history, not as the release throughput claim.

## Review

Review was deliberately adversarial and iterative:

- an initial independent review found the capacity-starved refinish assertion/cancellation defect
- Luna found an exact-cap natural-close rewind and fake-conductor EOS divergence
- after those fixes, Luna found the remaining marker/UTF-8 buffered-tail classification gap
- focused regressions were added for every defect
- final Luna code review: no actionable findings, confidence 0.94
- final Luna evidence review: no actionable findings, confidence 0.99
- final independent Sol review of `6ad7ee10`: no actionable findings, confidence 0.94

### DIFF
--- a/README.md
+++ b/README.md
@@ -634,11 +634,14 @@ Bounded sampled requests now retain normal speculative acceptance. The accepted
 round is observed before host emission; if committing it would consume reserved
 close/answer capacity, the engine retains only the prefix through the budget
 trip, re-finishes that sampled state, and installs the forced close before the
-next model decision. On one local RTX 3090 / width-8 / `Q27_BATCH=1` gate, the
-same 256-token bounded request improved from 32.4 t/s and 0.985 tokens/round to
-89.7 t/s and 2.639 tokens/round (2.77x); the unbounded control was 100.2 t/s and
-2.943 tokens/round. This is a single-prompt regression gate, not a universal
-throughput claim.
+next model decision. On the maintainer's matched RTX 5090 A/B
+(`temperature=0.7`, `top_p=0.95`, `max_tokens=700`, n=2 per arm), bounded
+throughput moved from 56.9 t/s on master to 147.9 t/s on this PR, while the
+unbounded control held at 143.1 -> 143.5 t/s (inside noise). Both bounded PR
+reps were exactly 147.9 t/s. The bounded arm used 275 rounds against 284
+unbounded, so its slight lead reflects the budget closing the trajectory early,
+not a faster kernel. This is a single-prompt gate, not a universal throughput
+claim.
 
 
 Why it defaults on: an 11-pack / 240-scenario A/B (BUILDLOG 2026-07-28, rescored
--- a/README.md
+++ b/README.md
@@ -625,16 +625,20 @@ Messages report a trip as `usage.reasoning_tokens` and
 `usage.reasoning_budget_exceeded`; Responses reports the same fields under
 `usage.output_tokens_details`.
 
-Greedy checks land after the accepted speculative round, so the reasoning count
-may exceed the configured limit by that round's accepted width. A natural close
-later in the same round ends normally rather than reporting a budget trip.
+Positive budget checks land after the accepted speculative round, so the
+reasoning count may exceed the configured limit by that round's accepted width.
+A natural close later in the same round ends normally rather than reporting a
+budget trip.
 
-Bounded sampled requests intentionally use one-token acceptance rounds so every
-accepted token is a coherent budget-observation point. This disables
-speculative acceptance for the bounded request: on the reviewed RTX 5090 run,
-the exact tip measured 56.9 t/s bounded versus 143.1 t/s unbounded, a 2.51x
-reduction. Unbounded sampled requests retain normal speculative width. Restoring
-speculative width without weakening transition semantics is follow-up work.
+Bounded sampled requests now retain normal speculative acceptance. The accepted
+round is observed before host emission; if committing it would consume reserved
+close/answer capacity, the engine retains only the prefix through the budget
+trip, re-finishes that sampled state, and installs the forced close before the
+next model decision. On one local RTX 3090 / width-8 / `Q27_BATCH=1` gate, the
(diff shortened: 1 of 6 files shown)


signalnine:
Matched A/B done, and the penalty is gone. No findings.

## The number you were waiting for

Same protocol as the 56.9 versus 143.1 pair currently in the README: your tip `6ad7ee10e055`, RTX 5090, `--request-think --think-budget 0`, one prompt, `temperature 0.7 top_p 0.95 max_tokens 700`, thinking enabled, n=2 per arm, the only difference between arms being whether the request carries `thinking.budget_tokens`.

| tree | bounded | unbounded | ratio |
|---|--:|--:|--:|
| master | 56.9 | 143.1 | 2.51x slower |
| this PR | **147.9** | 143.5 | 0.97x |

The bounded arm goes **56.9 to 147.9 t/s**, and the unbounded control does not move (143.1 to 143.5, inside noise). That control is the part that makes the result trustworthy: it says the bounded path is what changed and nothing else regressed to pay for it.

Both arms were deterministic across reps, 147.9 and 147.9 exactly.

Bounded is now marginally faster than unbounded rather than merely equal, which surprised me until I looked at the round counts: 275 bounded against 284 unbounded. That is the budget closing the block early and shortening the trajectory, not a kernel getting faster. Worth knowing so nobody reads the pair as speculative decode being better under a budget.

## The cost, stated

Bounded now overshoots. With `budget_tokens: 64` the response reported `reasoning_tokens: 66`, two over, with `reasoning_budget_exceeded` correctly true and a full 907-character answer after the close.

That is the trade working as designed rather than a defect. Exact cutoff and speculative width cannot both hold, and your README already says so: changing "Greedy checks" to "Positive budget checks" generalizes the overshoot note to cover this path, which was the one thing I was going to ask for. Two tokens is well inside an accepted round width.

## Gates on the same tip

Canonical `a2982c5197c627551b27d76a0a94b220` EXACT, all five sampling gates, and `fused_smoke` clean on solo/fused identity, conductor identity, A2 injected-error isolation, and both graph passes. The `Q27_BATCH=1` leg matters here because the retain-prefix-and-refinish path you describe runs through the conductor.

Please take my pair for the README rather than the 3090 width-8 figures, or carry both clearly labelled. The 56.9 that is published now was measured on this rig with this protocol, so 147.9 is its honest successor. Two numbers from two rigs invite the reader to compute a ratio across them that nobody measured.

manthedan:
Excellent — thanks for running the matched A/B and full gate set! I swapped the local 3090 figures in the README for your matched 5090 pair in `584b1125`: bounded 56.9 -> 147.9 t/s, while the unbounded control stayed flat at 143.1 -> 143.5 t/s.

I also captured n=2, the repeated exact 147.9 result, and the 275-versus-284 round counts so nobody reads the small bounded lead as a kernel speedup. The accepted-round overshoot contract stays right above it.

signalnine:
Checked the respin rather than assuming: `584b11254f4a` differs from the tip I measured only in `README.md`, with every file under `src/` and the `Makefile` byte-identical by object id. So the gates and the A/B both stand for the current tip without a re-run.

Thanks for carrying the round-count caveat into the README rather than just the favourable number. A reader seeing 147.9 against 143.5 would otherwise reasonably conclude speculative decode is faster under a budget, which is not what the pair says.

Ready from my side.

Issue #8, opened 2026-08-03
Metal merge-back: proposed order and review boundaries


manthedan:
Hi @signalnine — following your approval in #2, I completed a staged audit of the Metal
merge-back and identified the final cuts needed before any Metal-stack PR or
additional unordered standalone PR is opened. The explicitly invited reasoning
fix has merged as PR #9; its sampled-throughput follow-up is draft PR #10.

Before I open the remaining PRs, does the order and review boundary below work
for you? I want to give you explicit opportunities to reorder, narrow, or veto
each stage rather than create a pile of dependent PRs or make architectural
decisions on your behalf. The standalone fixes remain useful independently of
the Metal decision.

### Prepared scope

I did not use the raw fork diff as the proposal; it includes logs,
experiments, packaging, and project history. I rebuilt the contribution as a
serialized stack. The reasoning fix is now in current `master`. The three
remaining standalone branches must be refreshed onto that merge before send;
argmax still needs exact-tip GPU execution. Draft PR #10 final tip
`6ad7ee10e055` has exact host, local RTX 3090 canonical/sampling, rebuilt full-
kernel, live solo/conductor/concurrent/stream, and final Luna/Sol review evidence.
Its performance measurement predates the last boundary hardening and is
directional. The larger references below are provisional
staged replays and will be regenerated after their real parents merge and the
final cuts below are applied:

| stage | provisional audited delta |
|---|---:|
| backend seam, validation, and CUDA dtype guard | 13 files, +1,321/-30 |
| Metal core: backend, engine, shaders, CLI, tests, README | 14 files, +17,771/-4 |
| current combined serving replay: shared API/CUDA parity plus Metal HTTP, snapshots, recovery, and streaming | 28 files, +15,346/-613 |

The seam contains no Metal implementation. Its only `.cu` delta is four lines
in `src/device_model.cu`: two validation calls and their separators. `upload()`
validates each tensor before allocation/copy, while `upload_all()` validates
the whole model before entering its allocation loop. Runtime loading also
rejects missing, mis-sized, or malformed tensor payloads before either backend
can consume them; `DeviceModel` construction itself is unchanged.

The current replay is an audit reference, not a final PR stack. Before opening
the larger PRs I will:

- move the Metal-only KV-attribution instrument out of the backend-neutral seam
  and remove the generic inspector's CUDA-only Q8 embedding rule;
- remove uncalled benchmark/probe APIs, eager pipelines, and shaders from the
[...]

Issue #4, opened 2026-07-19
tool-call drift with large code-content arguments cannot be recovered (UN-RESCUED)


chaudhryfaisal:
# Summary 

Observed following errors on server and sessions stops on CC / Coding Agent 

## Occurrence 1
```bash
[q27] [timing] render 0.9ms encode 73.3ms (188028 chars -> 46554 toks)
[q27] [gen] prompt=46554 prefix_hit=46314 snap=46314 ckpt=-1
[q27] [gen-done] eos: 2709 tokens in 48.1s (56.3 t/s), n_max=32000
[q27] [req] rid=43 api=anth conv=79747cff011fa88 qw_ms=1 tok_ms=74 prompt=46554 hit=46314 ckpt=-1 pf=240 pf_ms=613 dec=2709 dec_ms=48087 cb_ms=32 rounds=790 tps=56.3 end=eos gw=0 yields=0 slot=0 t=283862 gch=68,106,151,119,978,0,0,0 gnh=115,263,251,178,615,0,0,0 glf=1354,1248,1097,978,0,0,0 gla=1280,1044,793,615,0,0,0 phd=7579.0 phv=39497.7 phs=2731 phwn=112,97,82,482,0,0,0 phwm=5507.5,5392.4,4112.7,24485.1,0.0,0.0,0.0 sfxm=968.8 sfxn=17 sfx=219,1267 bat=1.0,0
[q27] [drift] UN-RESCUED (ntools=27) intended tool call: {"name": "Write", "arguments": {"content": "// client.go — gRPC client for Balance BillingSummary RPC
[q27]
[q27] package main
[q27]
[q27] import (
[q27]   "context"
[q27]   "fmt"
[q27]   "time"
[q27]
[q27]   "google.golang.org/grpc"
[q27]   "google.golang.org/grpc/credentials"
[q27]   "google.golang.org/grpc/credentials/insecure"
[q27]   "google.golang.org/grpc/metadata"
[q27]   "google.golang.org/protobuf/types/known/timestamppb"
[q27]
[q27]   "balance/proto"
[q27] )
[q27]
[q27] const
[q27] [http] POST /v1/messages -> 200
   POST /v1/messages -> 200 OK  (duration: 49.6 s, execution: 49.3 s)
```

## Occurrence 2
```
start_server done
[q27] [http] GET /v1/models -> 200
[q27] [timing] render 2.1ms encode 76.4ms (264915 chars -> 67811 toks)
[...]

PR #11, opened 2026-08-04
stream_split: emit a boundary between adjacent tool calls

manthedan:
## Problem

`</tool_call><tool_call>` currently emits no separating segment. A consumer buffering one TOOL segment can therefore merge two calls, or lose the raw text of a malformed second call.

## Change

- Preserve an adjacent tool-call boundary across full and bytewise marker feeds.
- Emit a non-TOOL boundary before the next call begins.
- Preserve existing text, think, malformed-marker, and flush behavior.
- Include the focused stream-split regression in the default `all` build dependencies.

## Verification

Exact tip `50a82d92a273` on current upstream `e7e4b2554848`:

```text
adjacent/full: PASS
adjacent/bytewise: PASS
empty-think/full: PASS
empty-think/bytewise: PASS
nonempty-think/bytewise: PASS
text-between/bytewise: PASS
unfinished-empty-think/flush: PASS
```

`make build/test_stream_split && ./build/test_stream_split` passed. Final independent structured review found no actionable findings, confidence 0.96.

Independent standalone fix in the issue #8 approved sequence; it contains no Metal implementation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
 .PHONY: all clean
-all: build/inspect build/test_kernels build/q27 build/q27-server build/test_tokenizer build/test_depthctl build/test_toolconstrain
+all: build/inspect build/test_kernels build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
            src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h src/depthctl.h src/prefix_cache.h src/prefix_ram.h | build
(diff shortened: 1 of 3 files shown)


signalnine:
The fix is right and the new harness is better than what it replaces. One thing needs restoring before this goes in.

## The change itself

`tool_boundary` is the correct piece of state: set only when a TOOL closer returned to TEXT, cleared on any emitted non-TOOL payload, and consulted only where there genuinely is no segment to separate two calls. Threading it through `emit_head`'s new return value means a partial-marker flush also clears it, which is the case I would have gotten wrong. The empty-think handling in both `feed` and `flush` covers the variant where the separator has to be a THINK segment rather than a TEXT one.

Your seven assertions pass here, full and bytewise.

## What needs restoring

The rewrite drops all nine of master's regression cases:

```
stray close stripped
faisal multi: no </tool_call> in text
normal pair -> TOOL
normal pair TEXT=tail only
split stray tag held+stripped
think still routes
think text=ans
preseeded THINK: reasoning routes
preseeded THINK: answer after </think>
```

The first and fifth are the issue #4 regression this file was originally written for, a stray `</tool_call>` in TEXT being stripped rather than leaked, including the split-across-feeds variant. The second is a named incident. The last two matter more than they look: PR #9 just shipped a reasoning budget whose default scope is explicitly prompt-seeded `<think>` blocks, so those two cases now cover a shipped feature's precondition.

To be clear about what this is and is not: I compiled master's test file against **your** `stream_split.h` and all nine pass. The behaviour is intact, so this is a coverage regression rather than a correctness one. Nothing is broken; the guard rails against re-breaking it are gone.

I would not ask you to revert the harness. Comparing full segment vectors is strictly better than the old channel-concatenation approach, and bytewise plus whole-string is a real improvement. What I want is the nine cases ported into the new `expect(...)` form, so the file covers both the bug you fixed and the bugs it was already holding down.

Happy to do that port myself if you would rather not respin, just say so.

manthedan:
Thanks — addressed in `0d90548f1ff2`.

I ported all nine regressions from `master` into the exact segment-vector `expect(...)` harness, including the split-feed stray-close cases and prompt-preseeded THINK routing. The implementation is unchanged; this follow-up is test-only.

Verification:
- `make build/test_stream_split && ./build/test_stream_split`
- original seven boundary cases: PASS
- nine restored regression cases: PASS
- `git diff --check`: clean

Thanks for catching the coverage regression.

signalnine:
Verified on `0d90548f1ff2`. All sixteen pass, your seven boundary cases plus the nine restored, and `src/stream_split.h` hashes identical to the tip I reviewed, so this really is test-only.

Both stray-close variants and both preseeded-THINK cases are back by name. Ready from my side.

PR #12, opened 2026-08-04
fix: make tokenizer ownership explicit

manthedan:
## Problem

`Tokenizer` owns a heap-allocated PImpl but had no explicit ownership contract. A defaulted destructor leaked the implementation, implicit copying would duplicate the raw owner, and constructor failures could leave the tokenizer file open.

## Change

- make the PImpl a `std::unique_ptr` and define destruction where `Impl` is complete;
- delete copy construction/assignment and add `noexcept` moves;
- use RAII for the tokenizer file and checked reads for truncated artifacts;
- add a deterministic lifetime regression covering malformed construction, valid construction, moves, descriptor balance, and live implementation balance.

## Verification

- `make build/test_tokenizer && ./build/test_tokenizer --selftest`
- ASan build and `ASAN_OPTIONS=detect_leaks=0 ./build/test_tokenizer_asan --selftest`
- `leaks --atExit -- ./build/test_tokenizer --selftest`: 0 leaks
- final Codex branch review against `upstream/master`: no findings, correctness 0.98

Independent standalone fix in the issue #8 approved sequence. It contains no Metal implementation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ build/inspect: src/inspect.cpp src/loader.cpp src/loader.h | build
 	$(CXX) $(CXXFLAGS) src/inspect.cpp src/loader.cpp -o $@
 
 build/test_tokenizer: src/test_tokenizer.cpp src/tokenizer.cpp src/tokenizer.h src/api_common.h src/stream_split.h src/toolgram.h | build
-	$(CXX) $(CXXFLAGS) src/test_tokenizer.cpp src/tokenizer.cpp -o $@
+	$(CXX) $(CXXFLAGS) -DQ27_TOKENIZER_TESTING src/test_tokenizer.cpp src/tokenizer.cpp -o $@
 
 build/test_depthctl: tools/test_depthctl.cpp src/depthctl.h | build
 	$(CXX) $(CXXFLAGS) tools/test_depthctl.cpp -o $@
--- a/src/test_tokenizer.cpp
+++ b/src/test_tokenizer.cpp
@@ -3,11 +3,14 @@
 // cases.txt: alternating lines — text line, then space-separated reference ids.
 #include <algorithm>
 #include <atomic>
+#include <dirent.h>
+#include <unistd.h>
 #include <cstdio>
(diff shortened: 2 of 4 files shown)


signalnine:
No findings. Ready from my side.

The ownership treatment is complete rather than minimal, which is the right call for this one: a destructor alone would have left the copy and throwing-construction hazards you mention in the issue. Deleting copies, declaring the moves `noexcept`, and defining all three as `= default` in the translation unit where `Impl` is complete is exactly the shape a PImpl needs, and it is the part people usually get wrong by putting them in the header and getting an incomplete-type error they then work around badly.

`FileCloser` plus `read_exact` covers the other half. A throwing constructor now cannot leak the `FILE*`, and truncated payloads are rejected at construction rather than surfacing later as corrupt tokens.

The test-only instance counter behind `Q27_TOKENIZER_TESTING` is the detail I liked most. Counting live `Impl`s is the only way to assert the leak is actually gone rather than assert that a destructor exists, and gating it on a define kept it out of the production build instead of shipping an atomic nobody needs.

Verified here: dtor and both moves defined in `tokenizer.cpp` at lines 266 to 268, `build/test_tokenizer` and `build/inspect` both link (deleting the copy constructor does not break any existing call site), and the suite passes including your new `tokenizer failure lifetime` gate.

One note so it does not surprise you later, unrelated to this PR: `test_tokenizer` also carries a `billing-header cch normalize` case that has been failing on master since before any of this work, filed and not fixed. It sits after the `argc != 3` fixture check, so it does not run without `q27.tok cases.txt` and it did not run in my verification. It is header-normalization behaviour in `api_common.h` rather than anything you touched. Mentioning it only so a future full-fixture run does not get blamed on this change.

manthedan:
Thanks for the review and for flagging the pre-existing fixture-dependent billing-header failure. I reran `make build/test_tokenizer build/inspect && ./build/test_tokenizer --selftest`; both targets build and all file-independent selftests pass, including `tokenizer failure lifetime: PASS`. No source changes were needed.

PR #13, opened 2026-08-04
backend: add validated compute seam

manthedan:
## Problem

The Metal backend needs a backend-neutral execution contract, but that contract must land without hiding Metal implementation or weakening CUDA/model validation. Packed tensor payloads and sampling inputs also need one shared set of structural and numeric checks before either backend consumes them.

## Change

- add backend-neutral buffer, tensor, quantization, and compute interfaces under `src/backend.h`;
- centralize deterministic full-row and candidate sampling contracts in `src/sampling.h`, including NaN/+inf/no-finite rejection and mixed `-inf` masking;
- validate tensor payload sizes, offsets, scales, dtype declarations, and packed encodings during runtime loading and inspection;
- reject CUDA-incompatible embedding/layer dtypes before allocation or kernel launch;
- add CPU fixtures for loader, inspector, and sampling boundaries.

This PR contains no Metal implementation. Its CUDA runtime delta is limited to validation calls before upload/allocation.

## Verification

Exact tip `d0979cb4` on upstream `e7e4b2554848`:

- `make -B build/inspect build/test_loader_contracts build/test_sampling`
- `python3 tools/test_inspect.py ./build/inspect` — `inspect packed dtype fixtures: PASS`
- `./build/test_loader_contracts` — `loader CUDA dtype contracts: PASS`
- `./build/test_sampling` — `CPU sampling: PASS`
- `git diff --check upstream/master...HEAD`
- final Codex branch review: no accepted/actionable findings, correctness 0.91

Independent seam approved in issue #8; Metal backend, shaders, engine, and CLI/tests remain separate follow-up stages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,13 @@ build:
 
 build/inspect: src/inspect.cpp src/loader.cpp src/loader.h | build
 	$(CXX) $(CXXFLAGS) src/inspect.cpp src/loader.cpp -o $@
+build/test_loader_contracts: src/test_loader_contracts.cpp src/loader.cpp src/loader.h | build
+	$(CXX) $(CXXFLAGS) src/test_loader_contracts.cpp src/loader.cpp -o $@
 
-test-inspect: build/inspect
+
+test-inspect: build/inspect build/test_loader_contracts
 	python3 tools/test_inspect.py ./build/inspect
+	./build/test_loader_contracts
 
 build/test_sampling: src/test_sampling.cpp src/sampling.h | build
 	$(CXX) $(CXXFLAGS) src/test_sampling.cpp -o $@
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ build/mma16_bench: tools/mma16_bench.cu src/kernels.cu src/device_model.cu src/l
 	$(NVCC) $(NVCCFLAGS) tools/mma16_bench.cu src/kernels.cu src/device_model.cu src/loader.cpp -o $@
 
 build/test_kernels: src/test_kernels.cu src/kernels.cu src/prefill.cu src/blocks.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
-                    src/kernels.cuh src/prefill.cuh src/blocks.cuh src/spec3.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h | build
+                    src/kernels.cuh src/prefill.cuh src/blocks.cuh src/spec3.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h src/sampling.h | build
 	$(NVCC) $(NVCCFLAGS) src/test_kernels.cu src/kernels.cu src/prefill.cu src/blocks.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp -o $@
 
 
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_89,code=sm_89 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
-.PHONY: all clean
-all: build/inspect build/test_kernels build/q27 build/q27-server build/test_tokenizer build/test_depthctl build/test_toolconstrain
+.PHONY: all clean test-inspect
+all: build/inspect build/test_sampling build/test_kernels build/q27 build/q27-server build/test_tokenizer build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
            src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h src/depthctl.h src/prefix_cache.h src/prefix_ram.h | build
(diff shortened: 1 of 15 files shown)


signalnine:
The seam is a reasonable shape and the claims in #8 hold up. Two things I want resolved before it lands, one of which I think is important.

## Verified

The `.cu` delta really is four lines, and they really are just the two validation calls. The loader work underneath is the good kind of defensive: overflow checks on row and element counts, payload presence and size agreement, T2 and T3 encoding validity down to the 242 byte bound and the tail trit check, and alignment overflow on the tensor table.

Compatibility checked rather than assumed. I built your `inspect` and ran all thirteen `.q27` files I have here, every published tier of both the vanilla and fine-tuned models, and all thirteen still load. `loader CUDA dtype contracts` and `CPU sampling` both pass.

Keeping the `token_embd.weight must be Q8_G128` rule inside `validate_cuda_*` rather than the generic path is the right home for it, and matches what you said you would do in #8.

## 1. Two nucleus samplers, and nothing compares them

`sampling.h` is entirely new. There are no deletions anywhere else in the PR, so nothing was extracted; this is a second implementation standing alongside the device one that already ships in `blocks.cu` as `k_nucleus_d`, with its own argmax-always-in-nucleus guard and its own nucleus-mass computation.

Your `test_sampling.cpp` is genuinely good. The shuffled exact top-k over-set consuming RNG, tie reachability at the top-p cutoff, residual sampling never re-emitting an excluded draft: those are subtle and I would not have thought to write all of them. But every one of them tests the host implementation against its own specification. Nothing compares it to the device sampler whose behaviour is already gated, and whose nucleus-mass fix was specifically validated by empirical-accept matching p_served.

This is the failure mode this repo has already paid for twice. `turbo3_quant_group` is shared between the decode and prefill stores specifically so two writers cannot drift. The centroid tables in `turbo3.cuh` are macro-expanded into both the device `__constant__` array and the host mirror because a hand-retyped copy once produced 4096 silent mismatches. Two nucleus samplers with no cross-check is the same shape.

Concretely, the risk is that `top_p=0.95` comes to mean something slightly different on Metal than on CUDA, and nothing catches it. Exact token agreement is not the bar; the RNG streams differ and that is fine. What is comparable is deterministic: for given logits and a given top_p, which tokens are in the nucleus, and what are the renormalized probabilities over retained mass. That is testable against `k_nucleus_d` and it is the thing that would actually drift.

I would like that gate before this merges, or a stated reason it is not needed.

## 2. The seam already lags master on KV formats

`backend.h` makes turbo3 a first-class primitive: `kv_store_turbo3`, `attention_turbo3`, `turbo_wht`. There is no turbo5 equivalent, and turbo5k merged on 2026-08-01 as the Ampere default.

Today, adding a KV format costs a `kv_kind` case plus a store leg and a read leg. Under this interface it additionally costs new pure virtuals that every backend must implement or stub. That is a real tax on a part of the engine that has changed twice in the last month.

Two ways out and I do not have a strong preference. Either add the turbo5 primitives now so the seam matches master, or generalize the KV entry points so a format is a parameter rather than a method name and the vtable stops growing per format. The second is more work now and less later. Your call, but the current state where the interface cannot express a format that is already the default on one architecture should not be what lands.

## 3. Not a blocker, worth naming

Nothing in tree implements `ComputeBackend`. `engine.cuh` is a single line and the CUDA engine still calls its own kernels directly. So this interface is unvalidated by construction: its sufficiency rests on your out-of-tree Metal backend, and we find out whether the shape is right when Metal core lands on it.

I think that is still the right staging, and it is better than discovering the shape argument inside a 17k-line review. I just want it on the record that merging this is agreeing to an API on the strength of an implementation none of us can see yet, rather than a validated one.

## Nit

Several of the new lines have no space after commas or around operators, `matmul_quantized(const BackendTensor& weight,const BackendQuantized& x,` and `if(x_rows!=1)` among them. The rest of the tree is consistent the other way. Not worth a respin on its own, so fold it in whenever you next touch these files.

manthedan:
Thanks — addressed in `5ef3247d4787f460ef42ba6b2cad8c4321a6e33e`.

The seam now uses a backend-neutral `KvFormat` across the single-token and chunked KV store/attention contracts, while keeping caller-owned blocked-attention scratch explicit. I also added deterministic host/CUDA nucleus cross-checks for both retained support and served probabilities, and updated the Makefile dependency graph for the new sampling-header test input.

Verification:
- packed inspect fixtures: PASS
- loader CUDA dtype contracts: PASS
- CPU sampling regressions: PASS
- RunPod NVIDIA A40, CUDA 12.8 / `sm_86`, `build/test_kernels --sampling-only`: ALL PASS, including the new host/CUDA fixtures

No Metal implementation is included in this PR.

signalnine:
Both findings resolved on `5ef3247d4787`. Ready from my side.

**KV formats.** `turbo3` appears zero times in `backend.h` now, down from four, and `KvFormat` carries `TURBO5K = 4` in the same order as `KvKind`. Collapsing `kv_store_f16`/`kv_store_turbo3` into one `kv_store(..., KvFormat)` and the two attention entry points into `attention_causal` is the harder of the two options I offered and the right one: the vtable stops growing per format, so the next KV format costs an enum value rather than a virtual on every backend.

**The nucleus cross-check is exactly the gate I wanted**, and it is better than I asked for because it compares both dimensions separately rather than collapsing them into one number. Verified on Blackwell, RTX 5090, sm_120:

```
sample host/CUDA T=0.8 p=0.95 support        PASS  err 0.000e+00  tol 5e-01
sample host/CUDA T=0.8 p=0.95 probabilities  PASS  err 1.994e-08  tol 2e-05
sample host/CUDA T=1.5 p=0.95 support        PASS  err 0.000e+00  tol 5e-01
sample host/CUDA T=1.5 p=0.95 probabilities  PASS  err 1.129e-08  tol 2e-05
sample host/CUDA cutoff ties support         PASS  err 0.000e+00  tol 5e-01
sample host/CUDA cutoff ties probabilities   PASS  err 9.712e-09  tol 2e-05
```

Retained support matches exactly on all three, and the probability agreement sits three orders of magnitude inside tolerance. The cutoff-ties case is the one I would have worried about and it is the tightest of the three. Two nucleus implementations now have something holding them together, which was the whole concern.

Also reran the compatibility check: 13 of 13 shipped `.q27` tiers still load, and `loader CUDA dtype contracts` and `CPU sampling` both pass. `--sampling-only` needing no weights is a nice property, it let me run the Blackwell leg without taking the serving GPU down.

Still true and still fine: nothing in tree implements `ComputeBackend`, so the shape is validated by Metal core landing on it rather than by anything here.

PR #14, opened 2026-08-05
argmax: resolve exact and signed-zero ties to the lowest index

manthedan:
## Summary

- make exact-value ties resolve to the lowest index in both CUDA argmax paths
- treat `-0.0f` and `+0.0f` as IEEE-equal instead of ordering them by sign bit
- carry the winning index through intra-block and multi-block reductions
- add a focused CUDA regression target covering ordinary, signed-zero, fused/plain, and cross-block ties

This matches the existing CPU and Metal contract: equal maximum values select the lowest token index.

## Validation

Exact tip: `20b50692aee0b0877ad9e098cd939bf58c88bd1f`.

RunPod NVIDIA A40 (Ampere `sm_86`), CUDA 12.8.93:

```text
make -B build/test_argmax_tie build/test_kernels
./build/test_argmax_tie
  -0 then +0: plain=0 fused=0 expected=0 PASS
  +0 then -0: plain=0 fused=0 expected=0 PASS
./build/test_kernels /path/to/qwen36-27b-mtp-q4s.q27
  argmax tie -> lowest index (Metal/CPU rule) PASS
  ALL PASS
CANON_ARCH=sm86 tools/sampling_gate.sh /path/to/qwen36-27b-mtp-q4s.q27
  greedy canonical md5: 6894254e3b1a184ee3802771ddd59c2b
  SAMPLING GATES: ALL PASS
```

Final structured review reported no actionable findings. The maintainer's offered Blackwell leg can follow this clean Ampere-gated send.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
 .PHONY: all clean
-all: build/inspect build/test_kernels build/q27 build/q27-server build/test_tokenizer build/test_depthctl build/test_toolconstrain
+all: build/inspect build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
            src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h src/depthctl.h src/prefix_cache.h src/prefix_ram.h | build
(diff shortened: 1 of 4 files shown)


signalnine:
Blackwell leg done, no findings. Ready from my side.

## Gates on tip `20b50692aee0`, RTX 5090, sm_120, GPU otherwise idle

| gate | result |
|---|---|
| greedy canonical md5 | **`a2982c5197c627551b27d76a0a94b220` EXACT** |
| sampling gates 2-5 | ALL PASS |
| `test_kernels` | 353 PASS, 0 FAIL, including your new tie case |
| `test_argmax_tie` | both signed-zero orderings, plain and fused |

The canonical is the one that mattered here. A tie-break change is the single most likely thing to reroll a greedy trajectory, and it is the value you had no hardware to check. It did not move.

## On the change

Canonicalizing the sign bit before the order transform is the part I would have gotten wrong. IEEE equality already says the two zeros are one value, so without that the packed comparison would have silently ranked them by bit pattern and no amount of index tie-breaking would have helped. Inverting the index so a max reduction resolves to the lowest one, then undoing it in `am_unpack_idx`, keeps the whole thing to a two-line change in the hot path with no extra comparison.

The note that per-thread strided scans already keep their lowest index through strict `>` is the detail that makes it correct rather than accidentally correct, since the packed tie-break only has to handle the cross-thread and cross-block reduction.

Your `test_argmax_tie` is well aimed. `N = 32769` to span the multi-block reduction, and putting the second tied value at index 256 so the tie has to be resolved across blocks rather than inside one, is the case that would actually catch a wrong reduction. Testing both zero orderings catches the implementation that happens to work when the negative zero is encountered first.

One observation rather than a request: this makes the CUDA argmax match a rule that CPU and Metal already follow, so it is a convergence rather than a change of behaviour anyone was relying on. Worth saying in the commit message if it is not already, because a future reader finding a tie-break change on the decode path will want to know it moved toward a shared contract rather than away from one.

manthedan:
Fantastic — thanks for running the Blackwell leg! The exact canonical, 353/0 kernel result, and both signed-zero orderings close the one hardware gap I could not test locally.

And yes, exactly: this brings CUDA into the existing CPU/Metal lowest-index contract rather than changing intended behavior. Nothing else to change on my side.

PR #15, opened 2026-08-05
fix: clear pending boundary after stray tool close

manthedan:
## Problem

After a TOOL segment closes, `tool_boundary` records the pending segment boundary. If visible TEXT and a stray `</tool_call>` arrive in the same `feed()` call, the TEXT is emitted but the pending flag was left set. A later empty THINK block or tool opener could therefore emit an extra empty boundary segment.

## Fix

Clear `tool_boundary` when the visible prefix before a stray close is emitted, matching the other TEXT emission paths.

## Regression

`<tool_call>a</tool_call>x</tool_call><think></think>` now emits only TOOL `a` followed by TEXT `x`.

## Verification

- reproduced the merged behavior with a focused standalone probe
- `build/test_stream_split`: all 17 cases PASS, including the new regression
- `git diff --check`: clean

### DIFF
--- a/src/stream_split.h
+++ b/src/stream_split.h
@@ -64,7 +64,10 @@ struct StreamSplitter {
                         chan = TOOL;
                         continue;
                     }
-                    if (sc > 0) out.push_back({TEXT, hold.substr(0, sc)});
+                    if (sc > 0) {
+                        out.push_back({TEXT, hold.substr(0, sc)});
+                        tool_boundary = false;
+                    }
                     hold.erase(0, sc + strlen(C_CLOSE)); // strip stray close
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


signalnine:
Confirmed and reproduced. Ready from my side.

This one is mine to have missed. When I reviewed #11 I traced the `tool_boundary` clearing through the `<think>` opener branch, the `<tool_call>` opener branch and `emit_head`, and did not follow the stray-close branch, which is the one path that emits visible TEXT without clearing the flag.

Reproduced on merged master with a standalone probe on your input:

```
<tool_call>a</tool_call>x</tool_call><think></think>

master:  3 segments -> TOOL 'a', TEXT 'x', THINK '' (len 0)   <- spurious
PR #15:  2 segments -> TOOL 'a', TEXT 'x'
```

The extra empty THINK is exactly what the flag exists to emit when a separator is genuinely needed, arriving where one is not. A consumer buffering tool segments would see a flush it did not earn.

The fix is the right shape: clear on the visible-prefix emission, matching what the other TEXT paths already do, so the rule is now "any emitted non-TOOL payload clears the boundary" with no exceptions rather than three out of four.

`build/test_stream_split` is 17 for 17 here, the sixteen from before plus your new `text-before-stray clears boundary`.

PR #16, opened 2026-08-05
fix: require top-k for candidate sampling

manthedan:
## Problem

`top_k == 0` means full-vocabulary sampling, but the candidate helpers receive only a GPU-produced over-set. Treating that list as the whole distribution silently excludes the rest of the vocabulary and biases both ordinary sampling and speculative-rejection probabilities.

## Fix

Reject candidate sampling and candidate-built served distributions when `top_k` is disabled. Callers must use the full-logits path in that case; positive `top_k` candidate behavior is unchanged.

## Regression

Both candidate entry points now reject a finite incomplete candidate list with `top_k == 0`. Existing candidate/full-logit identity and nucleus-tie tests continue to cover positive `top_k` behavior.

## Verification

- reproduced the merged behavior with a focused standalone probe
- `build/test_sampling`: CPU sampling PASS
- RunPod NVIDIA A40, CUDA 12.8 / `sm_86`, `build/test_kernels --sampling-only`: ALL PASS
- `git diff --check`: clean

### DIFF
--- a/src/sampling.h
+++ b/src/sampling.h
@@ -101,6 +101,8 @@ inline uint32_t sample_candidates_cpu(const std::vector<float>& values,
     if(values.size()<count || indices.size()<count)
         throw std::runtime_error("q27: candidate list shorter than count");
     validate_logits(values.data(),count);
+    if(!p.top_k)
+        throw std::runtime_error("q27: candidate sampling requires top_k > 0");
     std::vector<uint32_t> order(count);
     for(uint32_t i=0;i<count;i++) order[i]=i;
     auto before=[&](uint32_t a,uint32_t b) {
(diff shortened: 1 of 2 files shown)


signalnine:
Ready from my side, and the reasoning holds.

A GPU-produced candidate over-set is not a distribution, so normalizing over it when `top_k == 0` quietly renormalizes the vocabulary down to whatever the extractor happened to return. That biases ordinary sampling and, worse, the speculative rejection probabilities, where a wrong served probability shows up as an acceptance rate that no longer matches `p_served` rather than as anything visibly broken.

I checked the blast radius before agreeing to a throw, since `SamplingParams` defaults `top_k` to 0 and a `runtime_error` on a serving path would be a 500 rather than a fix. `sampling.h` is included only by `src/test_kernels.cu`, and both candidate helpers are called only from `test_sampling.cpp`. Nothing on the serving path can reach either today, so the throw is safe to land.

Worth carrying forward rather than acting on now: when Metal core wires these up, a `top_k == 0` request has to take the full-logits fallback your description names. If a future caller forgets, this converts a silent bias into a surfaced exception, which is the better failure but still one somebody meets in production. A comment at the call site when it appears would be enough.

Changing the existing test from `top_k = 0` to `4` is correct rather than an accommodation: that test was exercising the invalid configuration, and the invalid case is now covered explicitly by the new rejection assertions.

Verified: `test_sampling` CPU sampling PASS, and the host/CUDA nucleus cross-check from #13 still passes on Blackwell since it shares this header, support exact and probabilities around 1e-8 against a 2e-5 tolerance.

PR #17, opened 2026-08-05
metal: add native compute backend primitives

manthedan:
## Summary

Add a standalone Apple Metal implementation of the merged `ComputeBackend` seam.

- F32/F16, Q8, Q4, and T2 projection primitives
- quantized activation and small-N matrix paths
- embeddings, normalization, RoPE, argmax, concatenation, and gating
- F16 and TURBO3 KV storage/attention, including chunk-causal execution
- GDN/DeltaNet decode and chunked primitives
- mmap-backed model views, command batching, optional per-dispatch profiling, and buffer/range hardening

This is the backend-only cut. It intentionally does not add the Metal engine or CLI yet, so the dependency boundary remains reviewable and the existing CUDA executable is unchanged.

## Verification

Run on Apple M4 against current `master`:

```text
make -B test-metal-backend
Metal matvec: OK
Metal decode primitives, FP16/turbo3 attention, GDN, and chunked prefill ops: OK
```

The same target also passes at the first commit in the two-commit stack, so both commits are independently buildable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_89,code=sm_89 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
-.PHONY: all clean test-inspect
+.PHONY: all clean test-inspect test-metal-backend
 all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
@@ -138,3 +138,30 @@ build/q27-server-w16: src/server.cu src/engine.cuh src/conductor.h src/blocks.cu
                       src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h | build
 	$(NVCC) $(NVCCFLAGS) -DQ27_W_MAX=16 -Xcompiler -pthread src/server.cu src/blocks.cu src/prefill.cu src/kernels.cu \
 	        src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp src/tokenizer.cpp -o $@
+
+# Native Metal backend primitives. Kept separate from the engine/CLI targets so
+# this dependency cut stays buildable and testable on its own.
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Darwin)
+METALFLAGS := $(CXXFLAGS) -Werror -fobjc-arc -pthread -I src/metal
+METALLIBS := -framework Foundation -framework Metal
+
+build/test-metal-backend: src/metal/test_metal.cpp src/metal/metal_backend.mm \
+                          src/metal/metal_backend.h src/metal/q27_kernels.metal \
+                          src/backend.h src/loader.cpp src/loader.h | build
+	$(CXX) $(METALFLAGS) src/metal/test_metal.cpp src/metal/metal_backend.mm \
+	       src/loader.cpp $(METALLIBS) -o $@
+
+build/test-metal-ops: src/metal/test_metal_ops.cpp src/metal/metal_backend.mm \
+                      src/metal/metal_backend.h src/metal/q27_kernels.metal \
+                      src/backend.h src/loader.cpp src/loader.h | build
+	$(CXX) $(METALFLAGS) src/metal/test_metal_ops.cpp src/metal/metal_backend.mm \
+	       src/loader.cpp $(METALLIBS) -o $@
+
+test-metal-backend: build/test-metal-backend build/test-metal-ops
+	./build/test-metal-backend
+	./build/test-metal-ops
+else
+test-metal-backend:
+	@echo "test-metal-backend requires macOS" >&2; exit 1
+endif
--- a/src/metal/metal_backend.h
+++ b/src/metal/metal_backend.h
@@ -1,0 +1,187 @@
+#pragma once
+
+#include "../backend.h"
+
+#include <memory>
+#include <string>
+
+namespace q27 {
+
+class MetalBackend final : public ComputeBackend {
+  public:
+    MetalBackend();
+    ~MetalBackend() override;
+    MetalBackend(const MetalBackend&) = delete;
+    MetalBackend& operator=(const MetalBackend&) = delete;
+
+    std::string name() const override;
+    std::shared_ptr<BackendBuffer> allocate(uint64_t bytes) override;
+    void write(BackendBuffer& dst, uint64_t offset, const void* src,
+               uint64_t bytes) override;
+    void read(const BackendBuffer& src, uint64_t offset, void* dst,
+              uint64_t bytes) override;
+    void zero(BackendBuffer& dst) override;
+    void copy(const BackendBuffer& src, uint64_t src_offset,
+              BackendBuffer& dst, uint64_t dst_offset, uint64_t bytes) override;
+    BackendTensor upload(const Tensor& tensor) override;
+    BackendTensor upload(const Model& model, const Tensor& tensor) override;
+    void begin_commands() override;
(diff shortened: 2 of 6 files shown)


signalnine:
Reviewed. No findings that block it, one documentation nit, and one thing I cannot do that you should know about before this merges.

## The review boundary held, and I checked rather than assumed

Every new file is under `src/metal/`. The only edit outside it is the Makefile, and the only deletion in the whole PR is the `.PHONY` line gaining a target. `all:` is untouched, the Metal rules are inside `ifeq ($(UNAME_S),Darwin)`, and the products are separate binaries rather than anything linked into `q27` or `q27-server`.

Verified on Linux at your tip: `make all` exits 0 and builds the same set as before, and `make test-metal-backend` correctly refuses with `requires macOS`. So this cannot move the CUDA canonical, and I did not need to re-run it to know that.

Decomposing this out of the 17,771-line reference was the right call and it shows. 5,601 lines with a single dependency direction is reviewable in a way the combined tree would not have been.

## The seam survived first contact, which is the answer to a caveat I filed

When I reviewed the seam I said merging it meant agreeing an API on the strength of an implementation none of us could see, and that we would find out whether the shape was right when Metal core landed on it. This is that moment, and it fits: `class MetalBackend final : public ComputeBackend` with `override` throughout, no additions to the interface required.

More pointed: `KvFormat` appears 48 times across all five variants including `TURBO5K`. Had you kept the per-format virtuals from the first revision, this PR would have needed `kv_store_turbo5` and `attention_turbo5` methods that did not exist, and we would be having a different conversation now. The generalization earned itself within a day.

## The no-copy upload is better than the contract required

`upload(const Model&, const Tensor&)` is the one place the seam's "default may copy, unified-memory backends may alias" wording had to cash out, and this is a good cashing out. Wrapping the whole page-rounded mapping once and binding tensors at offsets sidesteps per-tensor alignment entirely rather than failing on unaligned extents. The `weak_ptr` cache in `model_wraps` means one `MTLBuffer` per mapping no matter how many tensors bind into it. The residency-set deleter removing the allocation before a later `Model` destruction can unmap is the detail I would have expected to be missing.

The nit: `metal_backend.h` declares that override bare. `backend.h` carries the "the Model must then outlive the returned tensor" obligation, but a reader working in the Metal implementation sees a signature identical to the CUDA one, where `upload` copies and the obligation does not exist. One line at the concrete declaration would stop someone extending this from having to infer the aliasing from the `.mm`.

## What I cannot do

This is the first PR in the sequence I cannot gate. Everything above is structural: I can prove it does not touch CUDA, and I can read the Metal, but I cannot run it. The Metal legs need the M4 Pro, which is the contract we agreed in #8.

So this needs my no-CUDA-impact verification plus a maintainer run of `make test-metal-backend` on macOS before merge, and the second half is not mine. I would rather say that plainly than let a green-looking review imply more coverage than it has.

Two related notes for the slices after this one. The tests here are synthetic, constructing tensors in memory rather than loading an artifact, which is right for a primitives cut and keeps them runnable anywhere with a GPU. It does mean the q4s artifact contract from #8 is still owed by whichever slice brings the engine. And `-Werror` in `METALFLAGS` is a good default to have set from the first Metal PR rather than retrofitted later.

manthedan:
Added the concrete lifetime note at `MetalBackend::upload(const Model&, ...)` in `ac79f24c`: the model's mapped bytes are aliased, so the model must outlive the returned tensor.

Re-ran `make -B test-metal-backend` on the updated tip; both Metal executables pass on macOS.

**Update after the crossed replies:** the maintainer completed the target M4 Pro gate successfully against `1491384b` and verified that `1491384b..ac79f24c` is exactly the requested one-line header comment, so that evidence applies to the amended tip. No hardware leg remains pending for this primitives PR. The artifact-backed q4s gate remains with the engine slice, including recording which upload branch ran.

signalnine:
Metal gate passed on the target hardware. Both binaries build clean under `-Werror` and both suites pass:

```
Metal device: Apple M4 Pro
working set 17.8 GiB, max buffer 13.3 GiB, threadgroup memory 32.0 KiB
Metal matvec: OK
turbo3 synthetic attention: nrmse 0.10952, cosine 0.993990
Metal decode primitives, FP16/turbo3 attention, GDN, and chunked prefill ops: OK
```

Combined with the no-CUDA-impact verification above, that is #17 gated on both sides. Ready to merge from here.

## What the device banner tells us, which is the useful part

Two numbers came out of that run that neither of us had before, and one of them changes what the engine slice needs to test.

**Working set 17.8 GiB (about 19.1 GB).** This confirms the q4s guidance from #8 with a real number rather than my 75%-of-unified estimate. q4s at 15.46 GB leaves roughly 3.6 GB for KV and scratch. The default tier at 17.73 GB leaves about 1.4 GB, which is not the "will not run at all" I first said and not comfortable either. q4s remains the right target for the artifact contract.

**Max buffer 13.3 GiB (about 14.3 GB), and this is the one that matters.** Neither tier fits in a single `MTLBuffer`:

| tier | size | fits working set | fits ONE buffer | upload path taken |
|---|--:|---|---|---|
| q4s | 15.46 GB | yes | **no** | per-tensor views |
| default | 17.73 GB | yes | **no** | per-tensor views |

Your `upload(const Model&, const Tensor&)` already handles this correctly. The `mapped_size <= maxBufferLength` branch takes the whole-mapping wrap, and everything past it falls to the per-tensor page-aligned `wrap` lambda with its own bounds and limit checks. I looked for the fallback before raising this and it is there, so nothing is broken.

The consequence is about coverage rather than correctness. **On this hardware every real model takes the fallback path, and the fallback is the path your tests do not exercise.** `test_mmap_upload` builds a synthetic file of a few hundred bytes, so it goes through the whole-mapping branch every time. The branch that q4s will actually use on a 24 GB M4 Pro has no test behind it.

That is not a change I would ask for in this PR, since these are primitives and the artifact contract belongs to the engine slice. It is worth writing down now so the slice that loads a real model tests the per-tensor path deliberately rather than discovering it. A synthetic model larger than `maxBufferLength` is impractical as a fixture, so the realistic options are either a unit test that forces the fallback by comparing against a lowered threshold, or making the first real q4s load on the Mac an explicit gate with the path it took recorded.

Worth noting for the record that this is exactly what the review split was for. I could prove #17 does not touch CUDA from here, and could not have found this without the device.

signalnine:
The M4 Pro leg is not pending. I ran it on the M4 Pro and the banner is in the comment above, which crossed with yours by a few minutes.

## The amend checks out

Diffing your pre-push head `1491384b` against `ac79f24c` gives one line: the comment at the `upload` override. Nothing else moved.

So neither piece of evidence needs redoing. The Linux verification and the device run were both against `1491384b`, and the only difference is a comment in a header that Linux never compiles. The wording is right where it needs to be, at the point someone extending the class will meet it.

## Your base M4 is a useful second data point, if you print the banner

The number that decides which upload path a real model takes is `maxBufferLength`, and on the M4 Pro it came back at 13.3 GiB against a q4s artifact of 15.46 GB. Your machine has a different memory configuration, so it will report a different limit, and I would rather know it than guess.

Why it matters: if the limit tracks physical memory the way the M4 Pro figure suggests, then every Mac that can hold the q4s tier is a Mac whose single-buffer limit sits below it. The per-tensor fallback would not be an edge case for large machines. It would be the only path a real q27 model ever takes on Apple silicon, and the whole-mapping branch would be reachable only from the synthetic tests. That is a stronger claim than the one I made above, and I do not want to make it on one measurement. One line from your run settles it.

## What this adds to the engine slice contract

Nothing moves about who owes what. The q4s artifact gate is still the engine slice's, and these primitives should not carry it.

It does add a requirement: whichever slice loads the first real model should record which upload branch it took, not just that the load succeeded. If both our devices take the fallback, a green load that silently used the untested path is the failure that record exists to catch.

Ready to merge from my side.

manthedan:
Base M4 banner from exact tip `ac79f24c`:

```
Metal device: Apple M4
working set 17.8 GiB, max buffer 13.3 GiB, threadgroup memory 32.0 KiB
```

That matches the M4 Pro's 17.8 GiB working-set and 13.3 GiB single-buffer limits despite the different memory configuration. So both measured Apple-silicon configurations put the 15.46 GB q4s artifact above `maxBufferLength`; the per-tensor path is the expected real-model path, not an edge fallback.

I have carried the resulting engine-slice contract forward: the first artifact-backed load must record which upload branch was used, and the q4s gate must demonstrate the per-tensor path. No further primitives change is needed.

PR #18, opened 2026-08-05
metal: add q4s inference engine

manthedan:
## Summary

Add the q4s-only Metal inference engine on top of the merged native backend:

- validate the qwen35/q4s metadata, tensor names, shapes, and dtypes before allocation
- support serial and chunked prefill, FP16 or turbo3 KV, greedy/MTP generation, sampled generation, teacher-forced NLL, snapshots, reset, and streaming stop semantics
- route every KV operation through the merged `KvFormat` backend seam
- derive and gate the actual model mmap upload path against artifact size and the device limit; the 15.46 GB q4s artifact records `per-tensor` on Apple M4 because it exceeds the 13.32 GiB single-buffer limit
- keep `top_k == 0` sampling on the complete logits vector and cover that fallback in the artifact-backed contract gate

This is the engine slice only. It adds no CLI, server, CUDA path, canonical digest, benchmark/probe surface, or Bonsai-specific policy. The Makefile additions remain Darwin-only and do not alter `all`.

## Verification

Run on Apple M4 against `qwen36-27b-mtp-q4s.q27` (MD5 `7e5454e0c0ded717136ad3e42634ba25`):

- `make -B test-metal-backend` — backend and operator suites pass under `-Werror`
- `make build/test_sampling && ./build/test_sampling` — `CPU sampling: PASS`
- `make -B test-metal-contracts MODEL=qwen36-27b-mtp-q4s.q27` — independently derives the required upload path from the artifact size and device limit, records `model upload path: per-tensor (mapping 14.40 GiB, max buffer 13.32 GiB)`, and reports `Metal q4s engine contracts: OK`
- each commit compiles independently under the Metal `-Werror` flags
- `git diff --check upstream/master...HEAD` is clean

The exact incremental delta from merged `master` is six files, +1,457/-2.

## Files (6)

- Makefile [MODIFIED] (+26 -1)
- src/metal/metal_backend.h [MODIFIED] (+1 -0)
- src/metal/metal_backend.mm [MODIFIED] (+12 -1)
- src/metal/metal_engine.cpp [ADDED] (+1106 -0)
- src/metal/metal_engine.h [ADDED] (+169 -0)
- src/metal/test_metal_engine_contracts.cpp [ADDED] (+143 -0)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_89,code=sm_89 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
-.PHONY: all clean test-inspect test-metal-backend metal-engine
+.PHONY: all clean test-inspect test-metal-backend metal-engine test-metal-contracts
 all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
@@ -163,6 +163,19 @@ build/metal-engine.o: src/metal/metal_engine.cpp src/metal/metal_engine.h \
                       src/sampling.h src/suffixdraft.h | build
 	$(CXX) $(METALFLAGS) -c src/metal/metal_engine.cpp -o $@
 
+build/test_metal_engine_contracts: src/metal/test_metal_engine_contracts.cpp \
+                                   src/metal/metal_engine.cpp src/metal/metal_engine.h \
+                                   src/metal/metal_backend.mm src/metal/metal_backend.h \
+                                   src/metal/q27_kernels.metal src/backend.h \
+                                   src/loader.cpp src/loader.h src/sampling.h src/suffixdraft.h | build
+	$(CXX) $(METALFLAGS) src/metal/test_metal_engine_contracts.cpp \
+	       src/metal/metal_engine.cpp src/metal/metal_backend.mm src/loader.cpp \
+	       $(METALLIBS) -o $@
+
+test-metal-contracts: build/test_metal_engine_contracts
+	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
+	./build/test_metal_engine_contracts "$(MODEL)"
+
 metal-engine: build/metal-engine.o
 
 test-metal-backend: build/test-metal-backend build/test-metal-ops
@@ -173,4 +186,6 @@ test-metal-backend:
 	@echo "test-metal-backend requires macOS" >&2; exit 1
 metal-engine:
 	@echo "metal-engine requires macOS" >&2; exit 1
+test-metal-contracts:
+	@echo "test-metal-contracts requires macOS" >&2; exit 1
 endif
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_89,code=sm_89 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
-.PHONY: all clean test-inspect test-metal-backend
+.PHONY: all clean test-inspect test-metal-backend metal-engine
 all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
@@ -158,10 +158,19 @@ build/test-metal-ops: src/metal/test_metal_ops.cpp src/metal/metal_backend.mm \
 	$(CXX) $(METALFLAGS) src/metal/test_metal_ops.cpp src/metal/metal_backend.mm \
 	       src/loader.cpp $(METALLIBS) -o $@
 
+build/metal-engine.o: src/metal/metal_engine.cpp src/metal/metal_engine.h \
+                      src/metal/metal_backend.h src/backend.h src/loader.h \
+                      src/sampling.h src/suffixdraft.h | build
+	$(CXX) $(METALFLAGS) -c src/metal/metal_engine.cpp -o $@
+
+metal-engine: build/metal-engine.o
+
 test-metal-backend: build/test-metal-backend build/test-metal-ops
 	./build/test-metal-backend
 	./build/test-metal-ops
 else
 test-metal-backend:
 	@echo "test-metal-backend requires macOS" >&2; exit 1
+metal-engine:
+	@echo "metal-engine requires macOS" >&2; exit 1
 endif
--- a/Makefile
+++ b/Makefile
@@ -160,14 +160,15 @@ build/test-metal-ops: src/metal/test_metal_ops.cpp src/metal/metal_backend.mm \
 
 build/metal-engine.o: src/metal/metal_engine.cpp src/metal/metal_engine.h \
                       src/metal/metal_backend.h src/backend.h src/loader.h \
-                      src/sampling.h src/suffixdraft.h | build
+                      src/sampling.h src/suffixdraft.h third_party/json.hpp | build
 	$(CXX) $(METALFLAGS) -c src/metal/metal_engine.cpp -o $@
 
 build/test_metal_engine_contracts: src/metal/test_metal_engine_contracts.cpp \
                                    src/metal/metal_engine.cpp src/metal/metal_engine.h \
                                    src/metal/metal_backend.mm src/metal/metal_backend.h \
                                    src/metal/q27_kernels.metal src/backend.h \
-                                   src/loader.cpp src/loader.h src/sampling.h src/suffixdraft.h | build
+                                   src/loader.cpp src/loader.h src/sampling.h src/suffixdraft.h \
+                                   third_party/json.hpp | build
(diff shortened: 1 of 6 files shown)


signalnine:
Reviewed. One finding, nothing that blocks it, and two things from earlier reviews that this closes properly.

## No CUDA impact, verified

Only `Makefile` and `src/metal/` are touched. `all:` is untouched, the new rules sit inside the existing `ifeq ($(UNAME_S),Darwin)` guard, and both new targets have Linux stubs that refuse. Confirmed at your tip: `make all` exits 0 and builds the same set, `make metal-engine` and `make test-metal-contracts` both print "requires macOS" and exit non-zero.

So the canonical cannot move and I did not need to re-run it.

One thing to ignore in the GitHub diffstat: `docs/PORTING.md` and `README.md` show as changed because the branch is based on the `pr17` merge and master has since gained a docs commit. Those are master-only additions, not deletions by this PR. It merges clean.

## This closes the #17 coverage gap and the #16 forward-flag

Both were things I asked to be carried into this slice rather than fixed in the last one, and both are here.

The `top_k == 0` path is right. `generate_sampled_from_logits` and `stream_sampled_from_logits` both call `sample_logits_cpu(read_logits(), ...)`, so sampling never reaches the candidate helpers and the throw #16 added is unreachable from the engine. The comment says why rather than just what, which is the version that survives someone optimising it later.

Extracting the upload branch into `uses_per_tensor_upload()` and having `upload()` call it is better than what I suggested. The predicate and the branch cannot drift apart because they are the same expression, and a caller can now ask which path a model takes without inferring it from sizes.

## The finding: the path is recorded, not gated

`MetalEngine` prints `model upload path: per-tensor (mapping 14.40 GiB, max buffer 13.32 GiB)` at construction, and your PR body quotes it. But `test_metal_engine_contracts.cpp` never reads it. If a refactor made a 15.46 GB artifact take the whole-mapping branch, or the predicate drifted, the gate would still print `Metal q4s engine contracts: OK`.

That is a fair reading of what I asked for on #17, since I offered "record which path it took" as one of two routes. Having seen it land, recorded is weaker than I intended: it catches the regression only if a human reads the log on the run where it happens.

Everything needed to close it is already here. The test can derive the expectation itself from `engine.backend().max_buffer_length()` and the artifact size, then require the reported path to match, which also catches predicate drift rather than only a threshold change. That needs one accessor on the engine, since `model_` is private:

```cpp
bool used_per_tensor_upload() const { return per_tensor_upload_; }
```

The value is already computed at construction. Worth doing in this PR rather than deferring, because this is the slice where the artifact contract arrives and the fallback is the branch every real model on Apple silicon takes.

## `validate_architecture` is stricter than the CUDA path, and that is the right direction

This is the part I did not expect. It checks `general.architecture`, `quant_policy`, every shape against the compile-time constant, rope base and sections, nibble order, group sizes, the MTP layer count, and each tensor's name, dtype and shape before a single allocation.

The CUDA loader does not do this. `src/loader.cpp` validates the container thoroughly (magic, version, offsets, 256-byte alignment, payload extents, `token_embd` dtype) and stops there. Nothing checks `qwen35.block_count` against `N_LAYER = 64` or the embedding length against `N_EMBD`. A `.q27` built from a different checkpoint would be read at the wrong shape by the CUDA engine with no error, which is the failure mode this file makes impossible on Metal.

Not a change to make here. Filing it as a CUDA-side follow-up, because the asymmetry now runs the wrong way: the newer backend is the careful one.

## `attention_layer` hardcodes what CUDA reads from metadata, but fails closed

`static bool attention_layer(uint32_t layer) { return layer % 4 == 3; }` bakes in `full_attention_interval = 4`, where the CUDA engine reads `attn_layers` from the artifact and asks `is_attn_layer(il)`. So the interval is data-driven on one backend and compiled into the other.

I checked the artifact before raising it. `attn_layers` is `[3, 7, ..., 63, 64]`: the sixteen layers where `layer % 4 == 3`, plus 64 for the MTP layer, which you handle separately. The rule is exactly right for this checkpoint.

And it fails closed, which is what matters. `validate_architecture` asserts `qwen35.full_attention_interval == 4` and then walks the whole `attn_layers` array against the generated expectation, including the trailing 64. A checkpoint with a different layout throws at construction instead of executing the wrong graph. For a q4s-only engine slice that is the correct trade.

Worth knowing rather than acting on: a future checkpoint that moves the interval will load on CUDA and throw on Metal. That is my problem, not yours. `docs/PORTING.md` on master currently says the interval needs no engine edit, which stopped being true for the whole codebase the moment this merges, and I will add the Metal note.

## Two state fixes, checked rather than trusted

`7b53f17a` and `45df04ac` are the interesting commits, so I traced both.

The resident-state fix is real. `mtp_forward` was writing its rmsnorm output into `x1_`, and `x1_` is the resident hidden that `capture_state` snapshots alongside `logits_`. A draft that stopped before any lane committed would have left both corrupted. Routing the scratch through `h_` is safe here: no call site passes `h_` as the `hidden` argument, the draft loop starts at `x1_` and moves to `mtp_hidden_out_`, and `h_` is not part of the snapshot.

The streaming fix reorders so the sink and EOS decision happen before any recurrent, KV, hidden or logits mutation, which is the same close-edge discipline the CUDA conductor needed. I went looking for an underflow on `delivered - 1` in `encoded = exhausted ? delivered - 1 : delivered` and in `pending = predictions[delivered - 1]`, and it is not reachable: `offered = min(accepted + 1, remaining)` with `accepted + 1 >= 1` and `remaining >= 1` under the loop guard, so `offered >= 1` and the inner loop always delivers at least once unless it stopped, and both underflow sites are behind a `stop` return. The `accepted += delivered ? delivered - 1 : 0` guard covers the stop-on-first-token case.

The contract test is well aimed at these. Asserting that EOS and cancellation leave the resident logits byte-identical **and** that a retry reproduces the baseline continuation is stronger than asserting it did not crash, and the retry half is what would catch a partial state mutation that happened to leave the logits buffer alone.

## What I cannot do

Same boundary as #17. Everything above is structural. Running it needs the Mac, so this wants `make -B test-metal-backend` and `make -B test-metal-contracts MODEL=...q4s.q27` on the device before merge.

One observation on your run, since it answers a question I asked on #17. Your base M4 reports a 13.32 GiB maximum buffer and the M4 Pro reported 13.3 GiB. Two machines, the same number, so the guess that the limit tracks physical memory is not confirmed by this pair; it is consistent with both being 24 GB parts. The practical conclusion is unchanged and now has a second data point behind it: q4s does not fit one buffer on the hardware we have, and the per-tensor path is the only one a real model takes.

manthedan:
Closed the upload-path finding in `31e36e86`.

`MetalEngine` now retains the exact `uses_per_tensor_upload(model_)` result and exposes it through `used_per_tensor_upload()`. The artifact contract test independently reads the artifact size, applies page alignment, compares it with `max_buffer_length()`, and requires the retained path to match. A predicate/reporting drift now fails the gate instead of relying on log inspection.

Re-ran on Apple M4 at the amended tip:

- `make -B test-metal-contracts MODEL=.../qwen36-27b-mtp-q4s.q27` — records `per-tensor (mapping 14.40 GiB, max buffer 13.32 GiB)` and reports `Metal q4s engine contracts: OK`
- `make -B test-metal-backend` — both backend/operator suites pass under `-Werror`

No changes were made for the CUDA architecture-validation follow-up or the intentionally fail-closed q4s attention layout.

manthedan:
Side Note: In about a month I'll have a beefier machine to run these on, so the Metal branch can become more ambitious. Alibaba claims 3.8 27B will be released in days, so I am excited to see where that takes us as well.

signalnine:
Confirmed at `31e36e86`, and the shape you chose is better than the one I suggested.

I proposed asserting against the retained value. You made the test derive the expectation independently: it `stat`s the artifact, applies its own page rounding, and compares against `max_buffer_length()`, never calling `uses_per_tensor_upload()`. So the two sides can disagree, which is the whole point. Asserting against the predicate would have been a tautology that passed no matter which way the predicate drifted.

I checked the one thing that could have made that cross-check hollow. The test measures the file with `stat` while the engine's predicate measures `model.mapping_size()`, and those have to be the same quantity or the gate would fail on a correct build. `loader.cpp:196-203` sets `map_size_ = (size_t)st.st_size`, so the mapping is exactly the file. They agree.

Worth naming a property you may not have aimed for: this gate is symmetric. On hardware where q4s fits one buffer, `expected_per_tensor` comes back false and the same assertion gates the whole-mapping branch instead. It does not become dead weight on a machine where the fallback stops being the interesting path, which matters given your next note.

Both non-changes are right. The CUDA architecture validation was always mine, and it is done: `validate_arch()` now runs before the weight upload rather than after it, checks every shape key against its `engine.cuh` constant plus `token_embd`'s row count against `VOCAB`, and is deliberately tier-agnostic and deliberately silent on `full_attention_interval`, since the CUDA path reads `attn_layers` and genuinely does not assume an interval. Canonical is EXACT and a negative gate rejects five same-length metadata mutations with the offending key named. The fail-closed q4s attention layout stays as it is; for a q4s-only slice, refusing is the correct behaviour.

That is #18 gated on both sides. Ready to merge from here.

## On the two things in your side note

A better machine changes what the artifact contract can cover. The per-tensor path is the only one either of our devices exercises today, and a 48 GB or larger part would be the first hardware where `expected_per_tensor` flips and the whole-mapping branch gets real coverage. Worth running the contract gate there early rather than assuming the branch we have never executed still works.

On 3.8, there is already something in the repo for it. `docs/PORTING.md` on master tabulates every compile-time constant in `engine.cuh` against the `config.json` field it has to match, so a new checkpoint is decidable before downloading any weights, and it carries the recon I did on 2026-08-05 as a dated section. The short version: the evidence points at 3.8 staying on the `qwen3_5` graph. vLLM PR #50068 loads a Qwen3.8-Max FP8 checkpoint through `Qwen3_5ForCausalLM` and reads the same Gated DeltaNet config keys we do, the `Qwen3.5-27B` to `Qwen3.6-27B` config diff is metadata only with every shape identical, and `transformers/main` still has no `qwen3_6` or `qwen3_8` directory. Dense versus MoE for the 27B slot is genuinely unknown, and MoE is the one outcome that is a different engine rather than a port.

The file also now records the divergence this PR introduces, since it is the kind of thing that is invisible until it bites: the interval is data-driven on CUDA and computed on Metal, so a checkpoint that moves it loads on one backend and refuses on the other.

signalnine:
But yeah, very excited for 3.8 dropping soon and I aim to have first-day support

PR #19, opened 2026-08-06
metal: add q4s CLI and canonical gates

manthedan:
## Summary

Complete the final Metal core stage from #8 on top of merged PR #18:

- add a q4s Metal CLI with token-id or text prompts, validation-only mode, configurable context, MTP width, FP16/turbo3 KV, chunked/serial prefill, CPU sampling, and token-id dumps
- add Darwin-only `test-metal` and `test-metal-canonical` targets without changing `all`
- key published canonical digests by both architecture and artifact tier, preserving the existing CUDA digests, their measured provenance, and fail-closed behavior for unknown pairs
- add a Metal gate that pins the independently measured base-M4 and M4 Pro q4s trajectory, auto-selecting both under the measured `metal-m4:q4s` family canonical
- reject ambiguous `--tokens`/`--prompt` combinations even when either supplied value is empty
- distinguish missing `generated:` trajectories from actual canonical or seeded-output differences

This stage adds no server, shared HTTP/API surface, CUDA source/runtime change, benchmark/probe surface, or Bonsai-specific policy. The shared-tool changes are the architecture+tier canonical lookup and clearer failure classification in `tools/sampling_gate.sh`.

## Verification

Replayed and run on base Apple M4 from current merged `master` (`07e05905`) against `qwen36-27b-mtp-q4s.q27` (MD5 `7e5454e0c0ded717136ad3e42634ba25`):

- `make -B test-metal MODEL=... TOKENIZER=...` — backend/operator suites pass under `-Werror`; artifact validation passes; the CLI MTP smoke generates `Paris.` and its empty-value mutual-exclusion regressions reject before model loading
- `make -B test-metal-contracts MODEL=...` — `Metal q4s engine contracts: OK`
- `make -B test-metal-canonical MODEL=... TOKENIZER=...` — 128-token `metal-m4/q4s` MD5 `3a82b01053311807cc8bbaacdd8dcec7`; first 16 speculative/plain token IDs are byte-identical; `METAL CANONICAL GATES: ALL PASS`
- compiled the new CLI separately against unmodified upstream engine sources and the candidate tree on the same device; both 128-token ID files were byte-identical with MD5 `3a82b01053311807cc8bbaacdd8dcec7`
- smoke-tested turbo3 KV with MTP, serial prefill, and text-prompt sampling with `top_k=0`
- `make build/test_sampling && ./build/test_sampling` — `CPU sampling: PASS`
- shell syntax and known/unknown architecture-tier mapping checks pass
- hermetic sampling-gate checks distinguish an absent trajectory from differing seeded trajectories and preserve the successful gate path
- every commit builds `build/q27-metal` independently under the Metal `-Werror` flags
- `git diff --check upstream/master...HEAD` is clean

Independently derived by the maintainer on a 24 GB Apple M4 Pro at `5a80ce1`, using the same artifact bytes verified against `CHECKSUMS.md5`:

- the 128-token dump MD5 is `3a82b01053311807cc8bbaacdd8dcec7`, byte-identical to base M4
- forced `CANON_ARCH=metal-m4` reports `METAL CANONICAL GATES: ALL PASS`
- `make -B build/q27-metal` and `make -B test-metal-backend` pass under `-Werror`
- the real q4s artifact exercises `model upload path: per-tensor (mapping 14.40 GiB, max buffer 13.32 GiB)`, closing the previously unmeasured fallback path

At final tip `2aa95f1`, a full canonical run with an injected exact `Apple M4 Pro` brand string auto-selects `metal-m4:q4s`, reproduces MD5 `3a82b01053311807cc8bbaacdd8dcec7`, and reports both gate phases passing. Unknown Apple variants continue to fail closed.

The exact incremental delta is five files, +403/-46.

## Files (5)

- Makefile [MODIFIED] (+32 -1)
- src/metal/metal_cli.cpp [ADDED] (+197 -0)
- tools/canonical_md5.sh [ADDED] (+39 -0)
- tools/metal_canonical_gate.sh [ADDED] (+84 -0)
- tools/sampling_gate.sh [MODIFIED] (+51 -45)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_89,code=sm_89 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
-.PHONY: all clean test-inspect test-metal-backend metal-engine test-metal-contracts
+.PHONY: all clean test-inspect test-metal-backend metal-engine test-metal-contracts test-metal test-metal-canonical
 all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
@@ -179,6 +179,27 @@ test-metal-contracts: build/test_metal_engine_contracts
 
 metal-engine: build/metal-engine.o
 
+build/q27-metal: src/metal/metal_cli.cpp src/metal/metal_engine.cpp \
+                 src/metal/metal_backend.mm src/metal/metal_engine.h \
+                 src/metal/metal_backend.h src/metal/q27_kernels.metal \
+                 src/backend.h src/loader.cpp src/loader.h src/tokenizer.cpp \
+                 src/tokenizer.h src/sampling.h src/suffixdraft.h | build
+	$(CXX) $(METALFLAGS) src/metal/metal_cli.cpp src/metal/metal_engine.cpp \
+	       src/metal/metal_backend.mm src/loader.cpp src/tokenizer.cpp \
+	       $(METALLIBS) -o $@
+
+test-metal: test-metal-backend build/q27-metal
+	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
+	@test -n "$(TOKENIZER)" || { echo "set TOKENIZER=...tok" >&2; exit 2; }
+	./build/q27-metal "$(MODEL)" "$(TOKENIZER)" --validate-only
+	./build/q27-metal "$(MODEL)" "$(TOKENIZER)" --tokens 760,6511,314,9338,369 \
+	       -n 2 --ctx 16 --mtp 4 --dump-token-ids build/metal-smoke.ids
+
+test-metal-canonical: build/q27-metal
+	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
+	@test -n "$(TOKENIZER)" || { echo "set TOKENIZER=...tok" >&2; exit 2; }
+	tools/metal_canonical_gate.sh "$(MODEL)" "$(TOKENIZER)"
+
 test-metal-backend: build/test-metal-backend build/test-metal-ops
 	./build/test-metal-backend
 	./build/test-metal-ops
@@ -189,4 +210,10 @@ metal-engine:
 	@echo "metal-engine requires macOS" >&2; exit 1
 test-metal-contracts:
 	@echo "test-metal-contracts requires macOS" >&2; exit 1
+
+test-metal:
+	@echo "test-metal requires macOS" >&2; exit 1
+
+test-metal-canonical:
+	@echo "test-metal-canonical requires macOS" >&2; exit 1
 endif
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ build/q27-metal: src/metal/metal_cli.cpp src/metal/metal_engine.cpp \
                  src/metal/metal_backend.mm src/metal/metal_engine.h \
                  src/metal/metal_backend.h src/metal/q27_kernels.metal \
                  src/backend.h src/loader.cpp src/loader.h src/tokenizer.cpp \
-                 src/tokenizer.h src/sampling.h src/suffixdraft.h | build
+                 src/tokenizer.h src/sampling.h src/suffixdraft.h third_party/json.hpp | build
 	$(CXX) $(METALFLAGS) src/metal/metal_cli.cpp src/metal/metal_engine.cpp \
 	       src/metal/metal_backend.mm src/loader.cpp src/tokenizer.cpp \
 	       $(METALLIBS) -o $@
--- a/Makefile
+++ b/Makefile
@@ -191,6 +191,10 @@ build/q27-metal: src/metal/metal_cli.cpp src/metal/metal_engine.cpp \
 test-metal: test-metal-backend build/q27-metal
 	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
 	@test -n "$(TOKENIZER)" || { echo "set TOKENIZER=...tok" >&2; exit 2; }
+	@if ./build/q27-metal "$(MODEL)" "$(TOKENIZER)" --tokens 760 --prompt "" >/dev/null 2>&1; then \
+		echo "Metal CLI accepted --tokens with an empty --prompt" >&2; exit 1; fi
+	@if ./build/q27-metal "$(MODEL)" "$(TOKENIZER)" --tokens "" --prompt test >/dev/null 2>&1; then \
+		echo "Metal CLI accepted an empty --tokens with --prompt" >&2; exit 1; fi
 	./build/q27-metal "$(MODEL)" "$(TOKENIZER)" --validate-only
 	./build/q27-metal "$(MODEL)" "$(TOKENIZER)" --tokens 760,6511,314,9338,369 \
 	       -n 2 --ctx 16 --mtp 4 --dump-token-ids build/metal-smoke.ids
--- a/src/metal/metal_cli.cpp
+++ b/src/metal/metal_cli.cpp
@@ -75,7 +75,7 @@ int main(int argc, char** argv) {
     if (argc < 3) {
         fprintf(stderr,
                 "usage: %s model.q27 tokenizer.tok [--validate-only | --tokens id,id,... | --prompt text] "
-                "[-n count] [--ctx count] [--mtp width] [--prefill chunk|serial] "
+                "[-n count] [--ctx count] [--mtp width] [--kv fp16|turbo3] [--prefill chunk|serial] "
                 "[--temperature T --top-p P --top-k K --seed S] [--dump-token-ids file]\n",
                 argv[0]);
         return 1;
@@ -87,7 +87,7 @@ int main(int argc, char** argv) {
         std::string token_list, prompt_text, dump_token_ids;
         uint32_t count = 1, context = 128, mtp_width = 0;
         q27::SamplingParams sampling;
-        bool validate_only = false, serial_prefill = false;
+        bool validate_only = false, serial_prefill = false, turbo3_kv = false;
 
         for (int i = 3; i < argc; i++) {
             const std::string arg = argv[i];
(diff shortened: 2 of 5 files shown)


signalnine:
Reviewed and gated on the 5090. One thing to undo, two to add, and the arch-and-tier keying fixes a trap I left in this file.

I reviewed `31b88169` and you pushed `dc134b7` while I was running it. My CUDA results carry over unchanged, since the one-line delta is in `tools/metal_canonical_gate.sh` and never executes on the path I exercised, but the new commit is the first item below.

## Gated here

`sampling_gate.sh` at your tip, GPU otherwise idle:

| check | result |
|---|---|
| default env | **ALL PASS**, gate 1 `a2982c5197c627551b27d76a0a94b220` |
| `CANON_TIER=q4s` on the q4s artifact | gate 1 `f64e7c02252ca4c40cea62db662205e0` |
| unknown pair (`CANON_ARCH=sm89`) | refuses with guidance, exit 2 |
| explicit `CANON_MD5=` | still bypasses the table |
| invoked from another working directory | resolves, the source path is `$0`-relative |
| five `default` then `q4s` gate sequences | 5/5 clean |

And the lookup exhaustively without the GPU: all five architecture spellings the old `case` accepted still return `6894254e`, the four `sm120` spellings and the unset default return `a2982c51`, and `sm86:q4s`, `sm86:q6f`, `sm89:default`, `metal-m4:default`, `sm120:q6k` all refuse rather than guess. Nothing outside `Makefile`, `src/metal/` and `tools/` changes, `all:` is untouched, the Linux stubs refuse.

## Please revert dc134b7

```diff
-    "Apple M4") CANON_ARCH=metal-m4 ;;
+    "Apple M4"|"Apple M4 Pro") CANON_ARCH=metal-m4 ;;
```

This makes an M4 Pro auto-select the base-M4 digest, and I do not think it is what you want. Three reasons.

Your own description says the opposite: "The base-M4 digest is intentionally scoped to `metal-m4:q4s`. No M4 Pro digest is claimed or auto-selected." That was accurate at `31b88169` and is not accurate now, and the commit landed without a note, so anyone reviewing from the description sees a guard that no longer exists.

It is issue #7 on new silicon. The A40 run was lost because a digest from one architecture was presented as the gate for another. We learned that sm_86 is SM-count-independent by **deriving** it: the 3090 at 82 SMs and the A40 at 84 agreed bit for bit, and only then was one value published for both. Apple silicon may well behave the same across the M4 and M4 Pro, but that is a measurement nobody has taken. The exact-match guard you wrote was the version that refuses to assume it, and refusing was the right answer.

It also spends the check it was meant to protect. Gabe's 24 GB M4 Pro is the independent derivation leg we agreed in #8. Under this commit that run stops producing an independent value and starts comparing against yours, so if the two ever disagree it surfaces as a failing gate on his machine rather than as a second published canonical.

The good news is that this is cheap to settle rather than argue: **the M4 Pro is available now.** Running the gate there with `CANON_ARCH` forced gives a real digest. If it equals `3a82b010`, the alias is justified and lands with evidence behind it and a line in the table saying so. If it differs, `metal-m4-pro:q4s` becomes its own entry and the exact-match guard was protecting a real difference. Either outcome is better than either of us guessing, and it is one command.

## The arch-and-tier keying closes a bug of mine

Worth being explicit, since it is my bug you fixed. The old header listed the tier digests as free-floating comments with no architecture qualification, directly beneath a paragraph explaining that the canonical is per-architecture. Someone on a 3090 gating q4s would read that list, set `CANON_MD5=f64e7c02...`, gate an sm_86 run against an sm_120 digest, and read the mismatch as a regression in their own build. That is #7 again, one layer down. Keying by both and refusing unknown pairs makes it impossible to express, and promoting the tier digests from comments to resolvable entries means `CANON_TIER=q4s` now works where it used to require copying hex out of a comment.

## The reasoning left with the values

The thirty-line header is gone and the digests survive in the table, but that block was not documenting digests. It recorded three things learned from an incident:

- that a non-Blackwell run reporting `6894254e` **has reproduced its own canonical rather than failed to reproduce this one**, the exact sentence that would have saved the A40 run
- that where nothing is published, the same-device upstream/candidate differential **is** the correct gate and not a weaker substitute
- where `6894254e` came from: derived locally on our own 3090, matching the A40 report exactly, which is what established SM-count independence in the first place

Your error message keeps a compressed form of the second, which is the one that most needs to be at the point of failure. The third is the one I would most like back. A reader looking at `sm86:default` in the table cannot tell whether it was measured on hardware or copied out of a GitHub issue, and those two have very different standing when a gate later disagrees with it. Four or five lines in `canonical_md5.sh` next to the table, rather than back in the gate script.

## While running this: gate 2 cannot tell empty from different

Not caused by your PR, and worth fixing in it since you are already in this file.

I hit `FAIL seeded runs differ` once and spent a while chasing it as nondeterminism. It was not. I had two reproducers sharing the GPU, `cudaMalloc` failed in `device_model.cu:28`, and a run that produces **no** `generated:` line fails `[[ "$a" == "$b" && -n "$a" ]]` exactly like a run that produces a different one. Gate 1 in the same run reported `md5=d41d8cd98f00b204e9800998ecf8427e`, which is the md5 of the empty string presented as a trajectory digest.

Seeded sampling is fine: 20 of 20 identical single generations and 5 of 5 clean gate sequences. The gate's reporting is what misled me. A check that the `generated:` line is non-empty before hashing or comparing, with a distinct message, would have turned an hour into a minute, and it protects every future contributor whose card is a little too full.

## Ready when

Everything else is merge-ready from my side. The blocker is only dc134b7, and the fastest resolution is the M4 Pro derivation rather than a revert on principle.

manthedan:
Thanks — agreed. `dc134b7` assumed the result the independent M4 Pro leg was meant to measure.

Updated tip: `5a80ce1`.

- Reverted M4 Pro auto-selection; an `Apple M4 Pro` brand string again refuses with exit 2 until measured.
- Restored the sm_86 provenance beside the table: RTX 3090 derivation, independent A40 byte match, and what established SM-count independence.
- Added explicit non-empty trajectory checks before hashing or comparing. Missing output now reports `produced no generated trajectory`; genuinely different seeded outputs still report `seeded runs differ`. Gates 3 and 4 receive the same distinction.

Focused checks passed: shell syntax; hermetic empty-output, differing-output, and successful sampling-gate paths; and exact M4 Pro refusal. The successful path preserves the generated-line newline before hashing.

For the independent M4 Pro derivation at `5a80ce1`, with `MODEL` and `TOKENIZER` set:

```sh
make -B build/q27-metal && tmp=$(mktemp) && trap 'rm -f "$tmp"' EXIT && ./build/q27-metal "$MODEL" "$TOKENIZER" --tokens 760,6511,314,9338,369 -n 128 --ctx 256 --mtp 4 --dump-token-ids "$tmp" >/dev/null && md5 -q "$tmp"
```

If that reports `3a82b01053311807cc8bbaacdd8dcec7`, I will add the M4 Pro alias with the measurement recorded. If not, I will publish it as `metal-m4-pro:q4s` instead.

signalnine:
Derived on the 24 GB M4 Pro at `5a80ce1`:

```
3a82b01053311807cc8bbaacdd8dcec7
```

It matches your base-M4 value, so add the alias with the measurement recorded. `metal-m4-pro:q4s` is not needed.

Same artifact on both sides: `qwen36-27b-mtp-q4s.q27` md5 `7e5454e0c0ded717136ad3e42634ba25`, checked against `CHECKSUMS.md5` after the download, so this compares the same bytes rather than two things that happen to be named alike.

## I owe you a correction

I asked you to revert `dc134b7` and the alias in it was correct. Two separate things were true and I ran them together: the conclusion was right, and the evidence for it did not exist yet. The commit asserted cross-device digest equality, your PR description said no M4 Pro digest was claimed, and nothing had measured it. That is worth objecting to even when the guess turns out well, because the next assumption of that shape is the one that is wrong and looks identical beforehand. But my review led with "please revert" and it should have led with "please derive", which is what you proposed and what settled it in about a minute of GPU time.

## What the number establishes

The base M4 and the M4 Pro are different GPUs, more cores and more bandwidth, and they produce a bit-identical 128-token trajectory. That is the Apple-silicon analogue of the 3090-versus-A40 result you just restored to the table: two distinct parts, independently derived, agreeing byte for byte. Worth stating in the same comment as the provenance note, since it is the same kind of claim and now has the same kind of backing.

## Verified here at 5a80ce1

Your three changes, on hardware rather than by reading:

| check | result |
|---|---|
| auto-select on an `Apple M4 Pro` | refuses, exit 2, as intended |
| `metal_canonical_gate.sh` with `CANON_ARCH=metal-m4` | **METAL CANONICAL GATES: ALL PASS** |
| `make -B build/q27-metal`, `make -B test-metal-backend` | clean under `-Werror`, all ops OK |
| CUDA `sampling_gate.sh`, 5090 | **ALL PASS**, gate 1 `a2982c51...` |
| unknown pair `CANON_ARCH=sm89` | still exits 2 |

The empty-output distinction I checked against the failure that actually produced it, by pointing `BIN` at a stub that exits 0 and prints nothing:

```
FAIL canonical run produced no generated trajectory
FAIL seeded run produced no generated trajectory
FAIL seed comparison produced no generated trajectory
FAIL sampled/greedy comparison produced no generated trajectory
FAIL spec/plain run produced no generated trajectory
```

That is the message I wanted. The run that misled me was an out-of-memory `cudaMalloc` under my own GPU contention, and the old gate reported it as `seeded runs differ` while gate 1 printed the md5 of the empty string as though it were a trajectory. Both are now impossible to misread.

## One thing that fell out of the run

The gate reported `model upload path: per-tensor (mapping 14.40 GiB, max buffer 13.32 GiB)`.

That is the first time the per-tensor fallback has run against a real artifact on real hardware. It is the branch I flagged on #17 as having no coverage, the branch #18's contract test now asserts, and until today nothing had executed it with a 14 GiB mapping. It works. The coverage question from #17 is closed by measurement rather than by argument.

Not a request, but if you ever want a number for it: 1.83 tok/s at `-n 16` on this machine, which is a cold-start figure on an unoptimised path and not something I would put in a README.

## Ready

Merge-ready from my side once the alias lands with the measurement beside it. Everything else in this PR is gated on both architectures.

manthedan:
Measurement recorded and alias restored at `2aa95f1`.

- `Apple M4 Pro` now auto-selects `metal-m4:q4s`.
- The provenance beside the table records the independent 24 GB M4 Pro byte match, the shared artifact MD5 `7e5454e0c0ded717136ad3e42634ba25`, and the resulting family canonical.
- A full local canonical run with an injected exact `Apple M4 Pro` brand string selected `metal-m4/q4s`, reproduced `3a82b01053311807cc8bbaacdd8dcec7`, matched the 16-token plain prefix, and reported `METAL CANONICAL GATES: ALL PASS`.
- Unknown Apple variants still fail closed.

The distinction you made is right: the alias was a plausible conclusion before the run and a publishable canonical only after the independent derivation. I also added your real-artifact per-tensor result to the PR verification record; that closes the remaining coverage question from #17/#18.

Focused structured review is clean with no actionable findings.

PR #20, opened 2026-08-07
serving: restore shared API and CUDA parity

manthedan:
## Summary

- restore the shared OpenAI, Anthropic, and Responses serving behavior on current upstream after the Metal core series
- make wrapped and wrapper-less tool-call recovery stream-safe while keeping quoted, Markdown, HTML, and JSON-contained examples visible and non-executable
- preserve malformed/invalid output bytes, UTF-8 replacement semantics, tool-choice filtering, usage accounting, and truthful incomplete Responses lifecycles
- add ordinary optional API-key authentication plus exact-production auth/Responses integration coverage
- add explicit host gate targets without adding Metal serving, `ToolCallStreamer`, CUDA compiler-policy changes, or auth-rejection connection-close policy

## Scope

This is the shared API/CUDA serving stage from issue #8, replayed directly onto `3a15b1907de0`.

Exact source: `9d2f8663247aabffef72f268569a753276c0f494`
Tree: `01ae8352fabb315851772913ac9c9757861480e5`

13 files, +6,868/-634 versus the parent.

## Verification completed

On the exact source tip:

- pinned CUDA build passed with nvcc 12.8.93 / `sm_86`:
  - `make -B build/q27-server-w8 build/test_auth_integration`
- the exact production binary passed the model-backed Responses/auth matrix:
  - health exemption
  - missing, wrong, Bearer, and `x-api-key` authentication
  - streaming and non-streaming `/v1/responses`
  - token-cap and reasoning-budget truncation
  - legacy, current, and absent Codex headers
  - truthful terminal response/item status and `incomplete_details`
- extracted production handlers match their host fixture byte-for-byte:
  - `make check-chat-extract`
- host gates pass:
  - `build/test_stream_split`
  - `build/test_openai_bridge --selftest`
  - `build/test_tokenizer --selftest`
  - `build/test_chat_completions_integration`
  - `build/test_auth`
  - `build/test_auth_integration`
- `git diff --check` passes
- final structured review reported no accepted/actionable findings

## Live decode A/B

Matched A/B runs used the same RTX 3090, model and tokenizer artifacts, CUDA build flags, 128-token deterministic no-thinking response, and four sequential runs per leg. The first run was treated as warm-up; the table reports the median of the remaining three.

| Build | Median decode throughput |
|---|---:|
| upstream `3a15b1907de0` | 83.459 tokens/s |
| this branch `9d2f8663247a` | 83.479 tokens/s |
| delta | +0.024% |

The measured serving path is effectively unchanged at this precision.

## Model-free parser comparison

Median of seven matched runs against upstream `3a15b1907de0`:

| Workload | Upstream | This branch | Delta |
|---|---:|---:|---:|
| 8 KiB brace-heavy bare-call parse | 0.288 ms/request | 0.518 ms/request | +0.230 ms, 1.80x latency |
| 9 KiB 64-byte-chunk stream split | 0.131 ms/request | 0.022 ms/request | 5.86x faster |
| 128 KiB brace-heavy parser stress | 4.642 ms/request | 33.016 ms/request | 7.11x latency |

The 8 KiB case is the representative generated-output scale; the 128 KiB case records the worst synthetic brace-heavy regression rather than hiding it. The model-backed A/B above directly measures the live serving outcome.

## Files (13)

- Makefile [MODIFIED] (+26 -7)
- src/api_common.h [MODIFIED] (+2212 -115)
- src/markdown_lex.h [ADDED] (+771 -0)
- src/server.cu [MODIFIED] (+934 -393)
- src/stream_split.h [MODIFIED] (+48 -25)
- src/test_tokenizer.cpp [MODIFIED] (+9 -3)
- tools/extract_check.sh [MODIFIED] (+1 -0)
- tools/test_auth.cpp [MODIFIED] (+3 -1)
- tools/test_auth_integration.cpp [MODIFIED] (+26 -5)
- tools/test_chat_completions_integration.cpp [MODIFIED] (+746 -84)
- tools/test_openai_bridge.cpp [MODIFIED] (+1726 -1)
- tools/test_responses_integration.py [ADDED] (+219 -0)
- tools/test_stream_split.cpp [MODIFIED] (+147 -0)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
              -gencode arch=compute_89,code=sm_89 \
              -gencode arch=compute_120,code=sm_120 -Xcompiler -Wall
 
-.PHONY: all clean test-inspect test-metal-backend metal-engine test-metal-contracts test-metal test-metal-canonical
-all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_depthctl build/test_toolconstrain
+.PHONY: all clean test-inspect test-metal-backend metal-engine test-metal-contracts test-metal test-metal-canonical check-chat-extract check-responses-integration
+all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_openai_bridge build/test_chat_completions_integration build/test_depthctl build/test_toolconstrain
 
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
            src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h src/depthctl.h src/prefix_cache.h src/prefix_ram.h | build
@@ -30,11 +30,30 @@ test-inspect: build/inspect build/test_loader_contracts
 build/test_sampling: src/test_sampling.cpp src/sampling.h | build
 	$(CXX) $(CXXFLAGS) src/test_sampling.cpp -o $@
 
-build/test_tokenizer: src/test_tokenizer.cpp src/tokenizer.cpp src/tokenizer.h src/api_common.h src/stream_split.h src/toolgram.h | build
+build/test_tokenizer: src/test_tokenizer.cpp src/tokenizer.cpp src/tokenizer.h src/api_common.h src/stream_split.h src/markdown_lex.h src/toolgram.h | build
 	$(CXX) $(CXXFLAGS) -DQ27_TOKENIZER_TESTING src/test_tokenizer.cpp src/tokenizer.cpp -o $@
 
-build/test_stream_split: tools/test_stream_split.cpp src/stream_split.h | build
+build/test_stream_split: tools/test_stream_split.cpp src/stream_split.h src/markdown_lex.h | build
 	$(CXX) $(CXXFLAGS) -I src tools/test_stream_split.cpp -o $@
+
+build/test_openai_bridge: tools/test_openai_bridge.cpp src/api_common.h src/stream_split.h src/markdown_lex.h | build
+	$(CXX) $(CXXFLAGS) -I src tools/test_openai_bridge.cpp -o $@
+
+build/test_chat_completions_integration: tools/test_chat_completions_integration.cpp src/server.cu src/api_common.h src/toolconstrain.h src/toolgram.h src/stream_split.h src/markdown_lex.h src/tokenizer.h | build
+	$(CXX) $(CXXFLAGS) -I src tools/test_chat_completions_integration.cpp -o $@
+
+build/test_auth: tools/test_auth.cpp src/api_common.h | build
+	$(CXX) $(CXXFLAGS) -I src tools/test_auth.cpp -o $@
+
+build/test_auth_integration: tools/test_auth_integration.cpp src/api_common.h third_party/httplib.h | build
+	$(CXX) $(CXXFLAGS) -I src -I third_party -pthread tools/test_auth_integration.cpp -o $@
+
+check-chat-extract: tools/extract_check.sh src/server.cu tools/test_chat_completions_integration.cpp
+	./tools/extract_check.sh
+
+SERVER ?= build/q27-server
+check-responses-integration: $(SERVER) tools/test_responses_integration.py
+	python3 tools/test_responses_integration.py --server "$(SERVER)" --model "$(MODEL)" --tokenizer "$(TOKENIZER)"
 
 build/test_depthctl: tools/test_depthctl.cpp src/depthctl.h | build
 	$(CXX) $(CXXFLAGS) tools/test_depthctl.cpp -o $@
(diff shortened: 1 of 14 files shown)


signalnine:
Reviewed and gated on the 5090. Two findings, neither of which blocks the shape of this. The serving behaviour holds up under everything I could throw at it, and the one thing I want changed is a decay rule you already implemented for JSON and not for fences.

## Gated here

| gate | result |
|---|---|
| host suite, 11 binaries | all pass |
| `make check-chat-extract` | `build_prompt` / `prepare_anthropic_prompt` / `handle` byte-for-byte |
| splitter | 57 cases, full and bytewise |
| tool-drift corpus + drift + think-resolve | all pass, 9 fixtures, 15 modes |
| E2E serving, real server, Anthropic + OpenAI + Responses | **17 passed, 0 failed** |

The E2E covers API shapes, streaming `[DONE]` and `finish_reason`, greedy determinism, the whole reasoning-budget matrix including `budget 0`, tool calls with parseable arguments, and the over-context refusal.

`build/q27`'s translation units are untouched by this PR, so the CUDA canonical cannot move and I did not spend a run proving it.

Worth saying: #15's fix survives and is better than it was. Clearing the boundary once via `emit_text_head` before the marker branch means the rule is now "any emitted non-TOOL payload clears it" with no exceptions, rather than three paths out of four. That is the generalisation I asked for on #15 and did not expect to get for free.

## Finding 1: an unterminated fence silently disables tool calls for the rest of the response

The feature is right. A `<tool_call>` inside a fence, quote, or JSON string should stay visible, and echoed documentation should not be able to drive a client tool call. That is a real class of bug and the 57 splitter cases cover it well.

The mirror image is what I want to raise. An unterminated container never decays, so every later call in that response is swallowed. I probed it directly against both trees:

| input | master | this PR |
|---|---|---|
| bare tool call | fires | fires |
| call inside a **closed** fence | fires | inert (intended) |
| call after a **closed** fence | fires | fires |
| call after an **unterminated** fence | fires | **dropped** |
| call after an **unterminated** JSON string | fires | **dropped** |

Identical in whole-buffer and bytewise feeds.

What makes this a finding rather than a taste argument is that you already implemented the opposite principle one container type over. `malformed-json-then-tool` asserts that a malformed JSON container discards its invalid inner frames at an ancestor boundary so a later top-level call stays executable, and the comment in `markdown_lex.h` says exactly why: "so later top-level calls remain executable." Markdown fences have no equivalent path. A model that opens ` ```json ` and never closes it loses every subsequent tool call in that response, silently, where master would have fired them.

Two things bound it, and I want to be fair about both. The splitter is per-request (`StreamSplitter split` is a member of `ReasoningBudgetObserver`), so a stuck container cannot poison a session, only one response. And a model that opens a fence usually closes it before calling a tool. But truncation mid-fence is a documented drift mode, it is the reason `parse_bare_tool_calls` exists at all, and this is the one failure mode in the set that produces no diagnostic: the call simply never happens.

I am not asking for the feature back. I am asking whether a fence should decay the way malformed JSON does, and if the answer is deliberately no, for that to be stated so the next person reading the corpus does not think it is an oversight.

## Finding 2: the drift regression suite is not wired into the build

`test_tool_drift`, `test_tool_drift_corpus` and `test_think_resolve` have **zero** occurrences in the Makefile. No rule, not in `all:`. The binaries on this machine were dated 25 July and had never been rebuilt.

That is not something this PR introduced. It is something this PR makes dangerous, because it rewrites `api_common.h` by +2,212/-115 and those three are the regression suite for exactly that parser. Anyone running `make all` before and after would have seen the corpus "pass" without a single line of the new parser being compiled into it.

I rebuilt all three by hand against your tree and they pass, so this is a build-wiring gap and not a behavioural one. Rules mirroring `build/test_openai_bridge` would close it, and given you are already adding `test_openai_bridge` and `test_chat_completions_integration` to `all:` in this PR, it is the natural moment.

I would not have caught the state of those binaries except that my own first pass at the host gates ran against a half-built tree: I read a missing `EXIT=` sentinel as success and tested stale binaries, one of which did not exist. Everything in the table above is from a verified-clean rebuild. Three separate near-misses of that shape on this project is its own argument for wiring the suite into `make all`.

## Smaller notes

`tools/test_tool_drift_corpus.cpp` has a `printf` format warning (`%u` against an `int`) that surfaces once it is actually compiled with `-Wall -Wextra`. Cosmetic, but it will greet whoever adds the rule.

The `+0.024%` decode A/B is the right control to have run, and matching artifacts, flags and warm-up across legs is what makes it readable. No notes.

## Ready

Merge-ready from my side once Finding 2 is wired up, since that one is cheap and makes every future parser change safer. Finding 1 I would like an answer on rather than a patch: decay, or documented as deliberate.

manthedan:
Thanks — addressed in `64ec0c3b`.

**Finding 2:** `make all` now depends on `build/test_tool_drift`, `build/test_tool_drift_corpus`, and `build/test_think_resolve`, with the parser headers recorded as dependencies. I also split the corpus summary `printf`, removing the `%u`/`int` warning. I forced clean `-Wall -Wextra` rebuilds and ran all three binaries plus `test_stream_split`; all pass.

**Finding 1:** no decay is deliberate. `malformed-json-then-tool` recovers only when an actual matching ancestor closer establishes that parsing has returned to top level. An unterminated fence or JSON string has no equivalent trustworthy boundary. Decaying at `<tool_call>` itself would make content that is still syntactically fenced/string-contained executable, including truncated echoed or retrieved examples.

I added the rationale beside the fence-state decision and full-input plus bytewise regressions for both an unterminated fence and an unterminated JSON string. They remain visible/inert; a closed fence still permits the following call, and malformed JSON with a real ancestor boundary still recovers it.

Focused structured review is clean with no actionable findings.

signalnine:
Verified at `64ec0c3`. `make all` builds all three now, and I deleted the binaries first so the rule had to do the work rather than a stale artifact standing in for it. 61 splitter cases, 9 host gates green, and my own probe agrees with your new `unterminated-*` regressions on all four legs. The `markdown_lex.h` delta is comment-only, so the 17/17 E2E from `9d2f866` still describes this binary.

Your answer on the fence is better than what I proposed. I was reading the JSON recovery as the general rule and asking why fences were exempt. The distinction that actually matters is whether a matching ancestor closer establishes a trustworthy boundary, and an unterminated fence has none, so the only available decay point is the `<tool_call>` marker itself. Decaying there would make still-fenced truncated content executable, which is the hole the feature exists to close. Deliberate, documented, and pinned by regressions is the right resolution, and I would not want the decay I asked about.

One leftover, and it is the warning I predicted would greet whoever added the rule, just not the one I expected: `tools/test_tool_drift_corpus.cpp:10` trips `-Wcomment` now that the file actually compiles. It is a `//` line ending in a backslash, inside build instructions your new make rule supersedes. Cosmetic.

Merge-ready from my side.

PR #21, opened 2026-08-12
metal: add production serving stack

manthedan:
## Summary

- add the Metal HTTP server for OpenAI, Anthropic, and Responses APIs on top of the merged shared-serving contract
- add memory- and disk-backed prefix snapshots, atomic recovery, backend health handling, stream formatting, and focused serving-policy helpers
- preserve truthful incomplete Responses lifecycles, incremental tool-call streaming, optional one-principal API-key authentication, and the repository's single-operator security boundary
- publish the measured M4-family q4s canonical produced by the promoted serving shaders

## Scope

This is the Metal-serving stage separated from the Metal engine and shared API/CUDA serving work negotiated in issue #8. It is replayed directly onto the actual PR #20 merge.

Serving commit: `5013a57663e88285101beeec63f19d5b6faf1ec0`  
Diagnostic commit: `062ae1a133e3b81a5aff3c9091ce82a867182ff2`  
Exact tree: `b8364066d923db16617e6abbf672b2fe81a9f06f`  
Base: `c2b04865bde08f46db24774d3c454181440e2df2`

22 files, +11,234/-515 versus the parent. The second commit only adds the runtime Metal shader SHA-1 to the model-ready diagnostic so canonical reports identify the source actually compiled.

The branch removes Bonsai-specific artifact/execution support, dormant Metal probe APIs and kernels, private review provenance, and the no-op shared sampling refactor. Bonsai remains a separate follow-up proposal.

It excludes `docs/metal/`, activity logs, experiments, packaging, CUDA compiler-policy changes, request-body/aggregate-stop caps, bounded queue policy, browser filters, separate admin authorization, credential-scoped caches, extended existing-file privacy policy, and pre-response socket cancellation.

## Canonical publication

This serving stack intentionally changes Metal forward-pass numerics through promoted reciprocal-multiply activation quantization and revised shader arithmetic. The published `metal-m4:q4s` digest is:

```
f301095522174bdb99f75ec840ad1389
```

The exact corrected tip rebuilt locally with runtime shader SHA-1:

```
0292df9ee92de94a9af8d65dd916aabbf76342be
```

and produced `f301095522174bdb99f75ec840ad1389`, 56 speculation rounds, 127 drafts, and 72 accepted drafts (56.7%). The 16-token MTP/plain prefix remained byte-identical. The earlier contradictory `3a82b010...` report did not identify the shader loaded; this diagnostic removes that ambiguity, and the owner's report was correct.

## Review findings addressed

- every CUDA server target names `third_party/httplib.h` as a dependency; touching the header schedules all three rebuilds
- the vendored header records the q27 nonblocking peer-liveness patch so a vendor update cannot silently drop it
- the unrelated `src/sampling.h` refactor is absent
- Metal tool-constraint mask pools are request-scoped; slot assignment resets the allocator and invalidates the host-id map
- every teacher-forced NLL/logit path invalidates MTP cache state after advancing the main model

## Verification

On Apple M4, exact corrected source:

- clean `-Wall -Wextra -Werror` Metal CLI build: PASS
- `tools/metal_canonical_gate.sh MODEL TOKENIZER`: `f301095522174bdb99f75ec840ad1389`; runtime shader SHA-1 `0292df9ee92de94a9af8d65dd916aabbf76342be`; 16-token MTP/plain prefix byte-identical; `METAL CANONICAL GATES: ALL PASS`
- `build/test-metal-backend`: PASS
- `build/test-metal-ops`: PASS
- `touch third_party/httplib.h && make -n build/q27-server build/q27-server-w8 build/q27-server-w16`: all three compile commands emitted
- `git diff --check`: PASS
- remote branch OID matches `062ae1a133e3b81a5aff3c9091ce82a867182ff2`

The previous full exact-tree verification remains applicable to serving commit `5013a576...`; the only follow-up change is the startup diagnostic above. A concurrent rerun of the long engine-contract binary exceeded the one-hour harness timeout after reaching repeated real-model loads, so it is not claimed as a fresh pass here.

## Serving boundary

- loopback remains the default bind
- API keys remain optional and use one rotation set for every non-health route
- prefix caches remain content-keyed across that set
- production request failpoints are disabled; recovery failpoints exist only in the dedicated test build
- explicit memory budgets are honored by auto-context admission
- no header-triggered client exception changes Responses terminal semantics

Follows the completed decomposition in #8 and the shared-serving merge in #20.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ NVCCFLAGS ?= -O2 -std=c++17 -gencode arch=compute_86,code=sm_86 \
 
 .PHONY: all clean test-inspect test-metal-backend metal-engine test-metal-contracts test-metal test-metal-canonical check-chat-extract check-responses-integration
 all: build/inspect build/test_sampling build/test_kernels build/test_argmax_tie build/q27 build/q27-server build/test_tokenizer build/test_stream_split build/test_tool_drift build/test_tool_drift_corpus build/test_think_resolve build/test_openai_bridge build/test_chat_completions_integration build/test_depthctl build/test_toolconstrain
-
 build/q27: src/engine.cu src/engine.cuh src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp \
            src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/device_model.h src/loader.h src/cuda_common.h src/depthctl.h src/prefix_cache.h src/prefix_ram.h | build
 	$(NVCC) $(NVCCFLAGS) src/engine.cu src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp -o $@
@@ -63,7 +62,6 @@ check-chat-extract: tools/extract_check.sh src/server.cu tools/test_chat_complet
 SERVER ?= build/q27-server
 check-responses-integration: $(SERVER) tools/test_responses_integration.py
 	python3 tools/test_responses_integration.py --server "$(SERVER)" --model "$(MODEL)" --tokenizer "$(TOKENIZER)"
-
 build/test_depthctl: tools/test_depthctl.cpp src/depthctl.h | build
 	$(CXX) $(CXXFLAGS) tools/test_depthctl.cpp -o $@
 
@@ -93,7 +91,7 @@ build/test_argmax_tie: tools/test_argmax_tie.cu src/blocks.cu src/blocks.cuh | b
 build/q27-server: src/server.cu src/engine.cuh src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
                   src/device_model.cu src/loader.cpp src/tokenizer.cpp src/api_common.h src/stream_split.h src/markdown_lex.h \
                   src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/cuda_common.h src/toolgram.h \
-                  src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h | build
+                  src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h third_party/httplib.h | build
 	$(NVCC) $(NVCCFLAGS) -Xcompiler -pthread src/server.cu src/blocks.cu src/prefill.cu src/kernels.cu \
 	        src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp src/tokenizer.cpp -o $@
 
@@ -138,7 +136,7 @@ build/turbo5_test: tools/turbo5_test.cu src/turbo5.cuh src/turbo3.cuh | build
 build/q27-server-w8: src/server.cu src/engine.cuh src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
                      src/device_model.cu src/loader.cpp src/tokenizer.cpp src/api_common.h src/stream_split.h src/markdown_lex.h \
                      src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/cuda_common.h src/toolgram.h \
-                     src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h | build
+                     src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h third_party/httplib.h | build
 	$(NVCC) $(NVCCFLAGS) -DQ27_W_MAX=8 -Xcompiler -pthread src/server.cu src/blocks.cu src/prefill.cu src/kernels.cu \
 	        src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp src/tokenizer.cpp -o $@
 
@@ -163,7 +161,7 @@ build/fused_smoke: tools/fused_smoke.cu src/engine.cuh src/conductor.h src/prefi
 build/q27-server-w16: src/server.cu src/engine.cuh src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
                       src/device_model.cu src/loader.cpp src/tokenizer.cpp src/api_common.h src/stream_split.h src/markdown_lex.h \
                       src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/cuda_common.h src/toolgram.h \
-                      src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h | build
+                      src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h third_party/httplib.h | build
 	$(NVCC) $(NVCCFLAGS) -DQ27_W_MAX=16 -Xcompiler -pthread src/server.cu src/blocks.cu src/prefill.cu src/kernels.cu \
 	        src/spec3.cu src/vgemm.cu src/device_model.cu src/loader.cpp src/tokenizer.cpp -o $@
 
@@ -188,22 +186,24 @@ build/test-metal-ops: src/metal/test_metal_ops.cpp src/metal/metal_backend.mm \
 
 build/metal-engine.o: src/metal/metal_engine.cpp src/metal/metal_engine.h \
                       src/metal/metal_backend.h src/backend.h src/loader.h \
-                      src/sampling.h src/suffixdraft.h third_party/json.hpp | build
+                      src/sampling.h third_party/json.hpp | build
 	$(CXX) $(METALFLAGS) -c src/metal/metal_engine.cpp -o $@
 
 build/test_metal_engine_contracts: src/metal/test_metal_engine_contracts.cpp \
                                    src/metal/metal_engine.cpp src/metal/metal_engine.h \
                                    src/metal/metal_backend.mm src/metal/metal_backend.h \
                                    src/metal/q27_kernels.metal src/backend.h \
-                                   src/loader.cpp src/loader.h src/sampling.h src/suffixdraft.h \
+                                   src/loader.cpp src/loader.h src/sampling.h \
                                    third_party/json.hpp | build
 	$(CXX) $(METALFLAGS) src/metal/test_metal_engine_contracts.cpp \
 	       src/metal/metal_engine.cpp src/metal/metal_backend.mm src/loader.cpp \
 	       $(METALLIBS) -o $@
 
-test-metal-contracts: build/test_metal_engine_contracts
+test-metal-contracts: build/test_metal_engine_contracts build/test_metal_model_contracts build/test_metal_stream
 	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
 	./build/test_metal_engine_contracts "$(MODEL)"
+	./build/test_metal_model_contracts "$(MODEL)"
+	./build/test_metal_stream
 
 metal-engine: build/metal-engine.o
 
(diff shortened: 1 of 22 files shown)


signalnine:
One blocking finding: this moves the published Metal canonical and does not say so. Two smaller ones. The CUDA side is clean and I verified that rather than assuming it.

## Blocking: the Metal canonical changes

`make -B test-metal-canonical` on the 24 GB M4 Pro, your artifact, at `67e08b7`:

```
FAIL canonical changed: got f301095522174bdb99f75ec840ad1389
                       want 3a82b01053311807cc8bbaacdd8dcec7
```

I ran the control before reporting it. Merged master `c2b0486`, same machine, same artifact MD5 `7e5454e0...`, same command, immediately afterwards: `METAL CANONICAL GATES: ALL PASS`. So the move is this branch, not the environment.

Three things narrow what it is.

**The engine is self-consistent.** The gate aborts at step 1 and never reaches step 2, so I ran that comparison by hand. On this branch the MTP prefix and plain greedy agree byte-for-byte over 16 tokens. Speculation is not disagreeing with the serial path.

**Both trees are internally consistent and still disagree with each other**, which places the change in the forward pass rather than in speculation. Master's trajectory equals master's plain path, this branch's equals its own, and the two trajectories differ.

**The divergence is late, not structural.** The two 128-token trajectories are identical for the first 54 tokens, then 73 of the remaining positions differ. That is the signature of a small numerical difference accumulating until it flips one argmax, after which greedy decoding decorrelates completely. It is the same shape as the cross-architecture forking the sm_86 note in `canonical_md5.sh` describes, arriving from a code change instead of a different GPU.

`q27_kernels.metal` is +3,839/-185 here, so new or reordered arithmetic in the shaders is the obvious suspect.

**The new trajectory is consistent across both M4 parts, which is the good news.** I ran the same gate on a base Apple M4 (24 GB, separate machine, independently downloaded artifact, MD5 verified `7e5454e0...`) and it produced `f301095522174bdb99f75ec840ad1389` as well, with byte-identical speculation telemetry (56 rounds, 127 drafts, 72 accepted, 56.7%). So this is a code change with a stable result, not a device divergence, and the family-canonical property that `metal-m4` depends on survives it. Re-deriving is a publish, not an investigation.

I am not claiming the new trajectory is worse. I have no quality evidence either way, and it is visibly faster (3.45 vs 1.96 tok/s on the M4 Pro at `-n 128`). Either outcome is fine. What is not fine is that a 15,300-line PR moves a published canonical without mentioning it: the verification section lists model, engine, recovery, validation, request, auth, snapshot, stream and inference gates, and the canonical is absent from that list.

So, one of two things. If the shader work was meant to be numerics-preserving, this is a regression worth finding before merge. If it was not, the canonical needs re-deriving and publishing in `canonical_md5.sh` with the same provenance treatment the existing entries get, and the reason stating in the PR. Both derivations already exist above if you want them. The value of that digest is that it does not move quietly, and this is the first time in the series it has.

## The httplib patch is not wired into the build

`third_party/httplib.h` gains a nonblocking peer-disconnect probe (`is_alive`, `is_connection_alive`). The feature is reasonable and `src/metal/metal_server.cpp` is its only consumer, so the CUDA server gains dead code and no behaviour change.

The problem is the dependency graph. `third_party/httplib.h` is compiled into `build/q27-server` via `src/server.cu`, and it is **not** a prerequisite of that target:

```
touch third_party/httplib.h && make -n build/q27-server   ->  0 compile lines
touch src/api_common.h     && make -n build/q27-server   ->  1 compile line
```

So after this merges, anyone with an existing tree gets a `q27-server` that silently does not contain the patched header, and `make` reports success. `tools/test_auth_integration` already lists `third_party/httplib.h` correctly, so this is an inconsistency rather than a convention.

This is the same shape as the drift-suite gap in #20, and it is worth being blunt about why I keep finding these: a stale binary reporting a pass is the one failure mode that survives every other gate in the project.

Second, smaller point: patching a vendored library needs a record. Nothing in the tree marks `httplib.h` as locally modified, so the next vendor bump drops the probe silently and the Metal server loses disconnect detection with no compile error.

## `src/sampling.h` is a no-op refactor in shared CUDA code

Both argmax sites change from `min_element` with a "sorts before" comparator to `max_element` with an inverted one. I tested equivalence rather than reasoning about it: 200,000 randomised trials with a deliberately small value alphabet, 108,438 of which had a tied maximum, shuffled non-monotonic token ids. **Zero disagreements.**

So it is behaviour-preserving, which is the good news. It is also a change to a header compiled into `test_kernels` and `test_sampling` inside a Metal serving PR, for no behavioural gain, and shared-surface edits are exactly what forces a Metal PR to be gated on CUDA. Worth dropping from this branch unless it is load-bearing for something I have missed.

## Verified clean here

On Linux at `67e08b7`:

- `build/q27` and `build/q27-server` sources untouched, `all:` unchanged
- 12 host gates pass: stream_split, openai_bridge, tokenizer, chat_completions_integration, auth, auth_integration, tool_drift, tool_drift_corpus, think_resolve, depthctl, toolconstrain, sampling
- `make check-chat-extract` byte-for-byte on all three handlers
- all five Metal targets refuse correctly on Linux

## Metal gates, on a base M4

Everything else passed at `67e08b7`:

| gate | result |
|---|---|
| `make -B build/q27-metal` | clean under `-Werror` |
| `test-metal-backend` | matvec OK, turbo3 attention cosine 0.993990, all ops OK |
| `test-metal-contracts` | constrained MTP, stale-snapshot and chunk-logit contracts PASS; `Metal stream format: OK` |
| `test-metal-validation` | inspect packed-dtype fixtures PASS |

`test-metal-recovery` is the one I have not run, so the failpoint/recovery path is reviewed by reading only.

One incidental result worth recording, because it settles a question left open on #17. The base M4 reports `working set 17.8 GiB, max buffer 13.3 GiB`, identical to the 24 GB M4 Pro. Two different GPUs, same limits, so `maxBufferLength` is a function of installed memory rather than GPU tier. Both machines therefore take the per-tensor upload path on any real tier, exactly as the contract test asserts.

manthedan:
Resolved the blocking canonical regression by replaying the Metal-serving stage directly onto merged PR #20 (`c2b04865`) and restoring the q4s engine sources that the previous replay had overwritten.

Exact published tip: `67e08b7cb4cf15a128aa2567c8984f068be6885e` (tree `39c0ee4781654738355df6aee1be810725f38019`).

Evidence on the same 24 GB M4 Pro / same artifact:
- `make test-metal MODEL=... TOKENIZER=...`: PASS, including canonical MD5 `3a82b01053311807cc8bbaacdd8dcec7`
- `test-metal-contracts`: PASS
- `test-metal-recovery`: PASS
- `test-metal-validation`: PASS
- shared host/parser/tool/conductor/sampling gates: PASS
- production `build/q27-metal-server` launched with `Q27_METAL_FAIL_FINISH=1` and completed a real chat request normally, confirming the failpoint is isolated to the dedicated test build

The branch remains draft pending maintainer review. The prior failing tip is no longer published.

signalnine:
We disagree about the same tree, and I think there is a one-command way to find out why.

At `67e08b7cb4cf`, tree `39c0ee4781654738355df6aee1be810725f38019` (I checked the published commit's tree hash against the one you cite, so this is not a mispush), I get:

```
FAIL canonical changed: got  f301095522174bdb99f75ec840ad1389
                       want 3a82b01053311807cc8bbaacdd8dcec7
```

Three runs, two machines, every one a clean `make -B`:

| machine | result |
|---|---|
| 24 GB M4 Pro, macOS 26.5 | `f3010955...` |
| base Apple M4 24 GB, macOS 26.5.2, separate machine, independently downloaded artifact | `f3010955...` (twice) |
| merged master `c2b0486`, both machines, same artifact, immediately after | **PASS** at `3a82b010...` |

Same artifact MD5 `7e5454e0c0ded717136ad3e42634ba25` throughout. The master control rules out the environment.

## The diagnostic

The shaders compile at runtime from a file resolved through a search path (`Q27_METAL_SOURCE`, then `Q27_SHADER_PATH`, then a path derived from the executable, then a cwd-relative fallback), and **the canonical gate does not record which file it found**. So a disagreement like this cannot currently be attributed from the output alone.

The shader text is identifiable offline, because `source_sha1` is a plain SHA1 over the UTF-8 source:

| tree | `shasum src/metal/q27_kernels.metal` | canonical it yields |
|---|---|---|
| merged master `c2b0486` | `571a43f77740143df5314ff157bab2ad77f7e86d` | `3a82b010...` |
| this branch `67e08b7` | `0761aedd28f1afc88e52c88a04c718e38a6efbf4` | `f3010955...` |

On my base M4 at `HEAD=67e08b7`, `shasum src/metal/q27_kernels.metal` is `0761aedd...` and `Q27_METAL_SOURCE` is unset, so I can show I compiled this branch's shader.

**Could you run `shasum src/metal/q27_kernels.metal` at your tip?** If you get `0761aedd...` then we compiled the same text and the difference is somewhere else. If you get `571a43f7...`, your run picked up master's shader, which produces exactly the `3a82b010` you reported, and that would explain the whole thing without anything being wrong with your branch.

Two ways that happens without doing anything unusual: a `build/q27-metal` carrying a `Q27_SHADER_PATH` that points at another checkout, or a run whose resolved path lands in a different tree. Your evidence line is `make test-metal` without `-B`; all of mine used `-B`, which is also worth eliminating since it is free.

I withdraw one thing from my earlier framing. I was going to suggest the Metal toolchain version as the culprit, on the grounds that runtime compilation makes the canonical depend on the compiler. Then I read the compile options: `mathMode = MTLMathModeSafe` and `fastMathEnabled = NO`. With fast math off the compiler owes IEEE semantics, so cross-version drift is much less likely than I was about to claim, and the shader-identity question is the better one to ask first.

## Worth fixing regardless

`MetalBackend::shader_source_sha1()` already exists and `metal_server.cpp` uses it, but `metal_cli.cpp` does not, and the canonical gate runs the CLI. Printing that hash in the gate's output would have made this exchange a single comparison instead of a round trip, and it makes the digest self-describing: a canonical for a runtime-compiled shader should say which shader it compiled.

## Three fixes were lost in the revert

`129f3cb2` is no longer published. Independently of the canonical question it contained work worth keeping:

- `third_party/httplib.h` added as a prerequisite of `build/q27-server` and both `-w8`/`-w16` variants. I verified it: `touch third_party/httplib.h` went from 0 rebuild lines to 1. Without it the CUDA server silently goes stale against a patched header.
- a note in `httplib.h` recording that the vendored library is locally patched and naming `metal_server.cpp` as its consumer.
- the unrelated `src/sampling.h` argmax refactor reverted out of the branch.

## Coverage

Everything else on the Metal side passes at this tree on my machines: `-Werror` build, backend suites, engine and model contracts, stream format, inspect fixtures. `test-metal-recovery` you have run and I have not, so between us it is covered.

signalnine:
Re-derived at `062ae1a` on both parts, and the shader identity line does exactly what it needed to.

```
Metal q4s model ready on Apple M4     in 0.12 s (shader sha1 0292df9ee92de94a9af8d65dd916aabbf76342be)
Metal q4s model ready on Apple M4 Pro in 0.82 s (shader sha1 0292df9ee92de94a9af8d65dd916aabbf76342be)
  OK md5=f301095522174bdb99f75ec840ad1389
METAL CANONICAL GATES: ALL PASS
```

Both machines report a shader hash equal to `shasum src/metal/q27_kernels.metal` on the checkout, so each run now proves which shader text it compiled. That is the whole fix, and it is two lines.

Worth being explicit about why the re-run mattered rather than carrying the earlier numbers over: this tip has a **third** distinct shader (`0292df9e...`, against `0761aedd...` on the old tip and `571a43f7...` on master), and the digest in `canonical_md5.sh` was measured on the first of those. It still holds, but it was published ahead of a measurement on the tree carrying it, which is the same gap as before. With the hash printed, that is now self-checking rather than something either of us has to remember.

## Full Metal gate sweep, M4 Pro, `062ae1a`

| gate | result |
|---|---|
| `test-metal-canonical` | PASS, `f3010955...` |
| `test-metal` | decode primitives, FP16/turbo3 attention, GDN, chunked prefill OK; stream format OK; snapshot store metadata/shared-reuse/publication-locking PASS; q4s artifacts and Metal architecture OK |
| `test-metal-contracts` | engine contracts OK; constrained MTP, stale-snapshot and chunk-logit contracts PASS |
| `test-metal-validation` | inspect packed-dtype fixtures PASS |
| `test-metal-recovery` | **PASS** |

`test-metal-recovery` is the one I had not run before, so that closes my side. Base M4 independently passes canonical, contracts and validation.

## The three fixes are back, and I checked the Makefile one rather than reading it

- `third_party/httplib.h` is a prerequisite of `build/q27-server` and both width variants. `touch third_party/httplib.h` now yields 1 rebuild line for each of the three, against 0 before.
- the vendored-patch note is in the header, naming `metal_server.cpp` as the consumer.
- `src/sampling.h` is byte-identical to master, so the argmax refactor is out of the branch.

Outside `src/metal/`, the branch now touches only `Makefile`, `third_party/httplib.h`, `tools/canonical_md5.sh`, `tools/test_inspect.py`, `tools/test_responses_integration.py` and `tools/test_snapshot_store_shared.cpp`. That is a much smaller shared surface than the first cut.

## One loose end in the record

Your 04:02 comment states the resolution as restoring `3a82b010` at `67e08b7`, and the branch as it stands does the opposite: it accepts `f3010955` and updates the table with provenance, which is what the reverted-away `129f3cb2` did. I measured `67e08b7` three times across two machines and never reproduced `3a82b010` there, so I think that comment describes a tree that is no longer the proposal.

No action needed on the code. It is worth a line saying which resolution stands, because someone reading this thread in six months will otherwise find a comment asserting a digest the branch does not use.

Gated from my side.

manthedan:
Confirmed. The standing resolution is the corrected tip `062ae1a133e3b81a5aff3c9091ce82a867182ff2`: runtime shader SHA-1 `0292df9ee92de94a9af8d65dd916aabbf76342be`, canonical `f301095522174bdb99f75ec840ad1389`.

My 04:02 comment claiming `3a82b010...` at `67e08b7` is superseded and should not be treated as evidence for the current proposal. That run did not identify the runtime-resolved shader; the corrected CLI diagnostic and your two-machine re-derivation close the ambiguity. The PR body now records the accepted canonical and explicitly acknowledges that your report was correct.

Thanks for rerunning the full sweep and closing recovery. No further code action from your findings; the three reverted fixes are present as verified.

PR #22, opened 2026-08-13
metal: restore Bonsai model support

manthedan:
## Summary

- extend the merged Metal backend and engine with production T2 ternary and B1 binary execution for 64-block Bonsai artifacts
- validate Bonsai quantization policy/layout, run serial float-activation prefill, and fail closed for unsupported chunked-prefill and MTP paths
- restore the lossless PrismML Q2_0/Q1_0 repacker so the exact q27 Bonsai artifact is reproducible from the public Apache-2.0 source pack
- preserve merged official q4s serving behavior and keep Bonsai-only contract coverage optional

## Base and review scope

PR #21 merged as `c82f6642f9c5a1d75b175a11657229638cfed1de`. This branch is based directly on that merge.

Review only these three commits above merged master:

- `aaa6a23fa1e21dc2d8aad6f23465a351d97fe6e1` — Bonsai Metal execution and contracts
- `b91b98b46a7cb9ada82a30ddb2d37707c894351b` — reproducible public-source artifact conversion
- `1b88a20e29d7212ccdd33d4bda2566ae64ba3ae3` — address review findings

Exact source: `1b88a20e29d7212ccdd33d4bda2566ae64ba3ae3`  
Tree: `95b284889e5ff75a51fe77725f2426202759209b`  
Base: `c82f6642f9c5a1d75b175a11657229638cfed1de`

## Review resolution

- `BONSAI_MODEL` is optional again. `test-metal-contracts` always runs the official q4s engine, model, and stream coverage. It passes a Bonsai artifact to the engine-contract binary only when configured; otherwise the binary prints a Bonsai skip notice.
- The unrelated dspark/Q4_1 repack path is removed. `tools/repack.py` now adds only the T2 and B1 source-pack formats named by this PR; dspark input fails closed under the binary source-type contract rather than producing a second format.

The review-fix commit changes only the contract harness and repacker. It does not alter the Metal runtime implementation or shaders reviewed and exercised at `b91b98b4`.

## Reproducible artifact

The required source pack is public, ungated, and Apache-2.0:

- repository: `prism-ml/Ternary-Bonsai-27B-gguf`
- pinned revision: `abbae723028d71be674e71e1a71201a6f43fab22`
- file: `Ternary-Bonsai-27B-Q2_0.gguf`
- source size: `7165121600` bytes
- source MD5: `01493c0ade7d4fee416dfd874cdfcfd1`

```bash
hf download prism-ml/Ternary-Bonsai-27B-gguf \
  Ternary-Bonsai-27B-Q2_0.gguf \
  --revision abbae723028d71be674e71e1a71201a6f43fab22 \
  --local-dir models/ternary-bonsai-27b

python3 tools/repack.py \
  models/ternary-bonsai-27b/Ternary-Bonsai-27B-Q2_0.gguf \
  models/ternary-bonsai-27b/ternary-bonsai-27b-t2.q27
```

Expected output:

- size: `7154196992` bytes
- MD5: `8babb56d67b63f2308f883912a272bb3`

## Verification

Exact review-fix tip `1b88a20e`:

- `build/test_metal_engine_contracts` compiles cleanly with `-Wall -Wextra -Werror`
- the engine-contract binary accepts `OFFICIAL.q27` without a second argument
- `make test-metal-contracts MODEL=...` no longer hard-fails for an unset `BONSAI_MODEL`; it enters the official engine-contract leg
- `python3 -m py_compile tools/repack.py`: PASS
- `git diff --check`: PASS
- no dspark, Q4_1, or Q41 path remains in the proposal
- full repack on an independent Linux host in an isolated `gguf==0.19.0` environment: 851 tensors, 498 T2 tensors slot-verified, 26.89 B ternary weights, 29.7% zeros, output size `7154196992`, MD5 `8babb56d67b63f2308f883912a272bb3`

Metal runtime evidence at unchanged runtime tip `b91b98b4`:

- clean Apple M4 `-Werror` CLI/backend/ops builds
- backend and ops suites pass
- merged official q4s `--validate-only` passes
- maintainer independently confirmed `test-metal-canonical`, `test-metal`, and `test-metal-validation`, including canonical `f301095522174bdb99f75ec840ad1389`
- deterministic Bonsai smoke emits `Paris.` with token-file MD5 `ddae10a2e232f528e20b66dd4e7ad20e`

Generated model artifacts are not committed to the repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,8 @@ build/test_metal_engine_contracts: src/metal/test_metal_engine_contracts.cpp \
 
 test-metal-contracts: build/test_metal_engine_contracts build/test_metal_model_contracts build/test_metal_stream
 	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
-	./build/test_metal_engine_contracts "$(MODEL)"
+	@test -n "$(BONSAI_MODEL)" || { echo "set BONSAI_MODEL=...bonsai-t2.q27" >&2; exit 2; }
+	./build/test_metal_engine_contracts "$(MODEL)" "$(BONSAI_MODEL)"
 	./build/test_metal_model_contracts "$(MODEL)"
 	./build/test_metal_stream
 
--- a/Makefile
+++ b/Makefile
@@ -201,8 +201,11 @@ build/test_metal_engine_contracts: src/metal/test_metal_engine_contracts.cpp \
 
 test-metal-contracts: build/test_metal_engine_contracts build/test_metal_model_contracts build/test_metal_stream
 	@test -n "$(MODEL)" || { echo "set MODEL=...q4s.q27" >&2; exit 2; }
-	@test -n "$(BONSAI_MODEL)" || { echo "set BONSAI_MODEL=...bonsai-t2.q27" >&2; exit 2; }
-	./build/test_metal_engine_contracts "$(MODEL)" "$(BONSAI_MODEL)"
+	@if test -n "$(BONSAI_MODEL)"; then \
+		./build/test_metal_engine_contracts "$(MODEL)" "$(BONSAI_MODEL)"; \
+	else \
+		./build/test_metal_engine_contracts "$(MODEL)"; \
+	fi
 	./build/test_metal_model_contracts "$(MODEL)"
 	./build/test_metal_stream
 
--- a/src/metal/metal_backend.h
+++ b/src/metal/metal_backend.h
@@ -18,6 +18,7 @@ class MetalBackend final : public ComputeBackend {
     static const char* shader_abi_tag();
     std::string shader_source_sha1() const;
     bool gemm_half_enabled() const;
+    bool gemm_half_q4_enabled() const;
     uint32_t gqa_tile() const;
     uint32_t gqa_block() const;
     uint32_t gqa_threshold() const;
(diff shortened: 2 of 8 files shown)


manthedan:
Full public-source artifact reproduction completed on exact tip `f3f518314ac974f430bb4214beb2f6f540de9de6`.

```text
repacked 851 tensors: 107.58 GB f32-equiv -> 7.15 GB
T2 slot verification passed on 498 tensors (26.89 B ternary weights, 29.7% zeros overall)
MD5: 8babb56d67b63f2308f883912a272bb3
cmp: byte-identical to the contract artifact
```

Source: `prism-ml/Ternary-Bonsai-27B-gguf@abbae723028d71be674e71e1a71201a6f43fab22`, `Ternary-Bonsai-27B-Q2_0.gguf`, source MD5 `01493c0ade7d4fee416dfd874cdfcfd1`. The output is 7,154,196,992 bytes.

manthedan:
PR #21 merged as `c82f6642f9c5a1d75b175a11657229638cfed1de`. I replayed the two Bonsai commits directly onto that merge and force-with-lease updated this branch to `b91b98b46a7cb9ada82a30ddb2d37707c894351b`.

The eight-file proposal diff is content-identical to the pre-merge tip; only ancestry changed. Post-merge Apple M4 checks passed: clean `-Werror` builds, backend and ops suites, official q4s `--validate-only`, custom type-42 repack smoke, and deterministic Bonsai output `Paris.` with token-file MD5 `ddae10a2e232f528e20b66dd4e7ad20e`.

signalnine:
Reviewed the incremental delta over #21's tip, since re-reading the shared 11k lines would not tell either of us anything new. Two findings, and the q4s preservation claim holds where I could test it.

Scope I actually looked at: `git diff c82f6642..b91b98b4` (verified content-identical to the pre-merge `062ae1a..f3f5183`), which is +4,358/-175 across 8 files, of which `tools/repack.py` is the only one outside `src/metal/`.

## The q4s claim holds, and the shader chain now explains itself

`test-metal-canonical` on a base M4, your artifact:

```
Metal q4s model ready on Apple M4 in 0.13 s (shader sha1 0761aedd28f1afc88e52c88a04c718e38a6efbf4)
  OK md5=f301095522174bdb99f75ec840ad1389
METAL CANONICAL GATES: ALL PASS
```

So +2,531 lines of T2/B1 shader do not perturb the q4s trajectory. That is the claim worth checking and it survives.

The hash also resolves something from the #21 thread. `0761aedd` is exactly the shader from the original `67e08b7` tip, and `0292df9e` is #21's corrected one. Both produce `f3010955`. So the original #21 already carried the Bonsai kernels, `062ae1a` stripped them per the #8 cut, and this PR re-proposes them, with the q4s digest unchanged in both directions. The shader-hash line you added made that legible in one run instead of three.

`test-metal` and `test-metal-validation` also pass here.

## Finding 1: `test-metal-contracts` now hard-requires a Bonsai artifact

```
set BONSAI_MODEL=...bonsai-t2.q27
make: *** [test-metal-contracts] Error 2
```

The target gained `@test -n "$(BONSAI_MODEL)" || ... exit 2` ahead of every recipe line, so the whole target fails without a Bonsai model, including `test_metal_model_contracts` and `test_metal_stream`, which do not take `BONSAI_MODEL` and have nothing to do with Bonsai.

The effect is that q4s contract coverage becomes conditional on an optional artifact. Anyone with only the official q4s tier, which is everyone following the #8 contract, loses a gate they had at #21, and loses it as a hard failure rather than a skip.

Running the Bonsai leg when `BONSAI_MODEL` is set and skipping it with a printed notice otherwise keeps both audiences. The rest of the target should not depend on it at all.

## Finding 2: `repack.py` also gains a dspark pack type, unmentioned

The summary describes T2 ternary and B1 binary Bonsai support. `repack.py` additionally gains a complete `dspark` path: `DTYPE_Q41 = 7`, a `dspark-q41-v1` quant policy, a `dspark.` metadata prefix, its own tensor-type validation, and a rule that everything not byte-copied widens to F32 so the name policy cannot requantize it. The PR body mentions dspark zero times.

This is a shared CUDA-side tool, and `repack.py` is how every `.q27` artifact including the published tiers gets made, so a second undocumented format is worth naming in a PR titled "restore Bonsai model support". It is also the same shape as the #8 cut that separated Bonsai out in the first place.

To be clear about risk rather than scope: **the existing path is byte-identical, and I proved it rather than reading it.** I repacked q4s from the original `Qwen3.6-27B-MTP-BF16.gguf` with this branch's `repack.py` (`--q4-head --tag q4s-v1`) and `cmp` reports no difference against the published `qwen36-27b-mtp-q4s.q27`, MD5 `7e5454e0c0ded717136ad3e42634ba25`. A master control from the same input reproduces it too, so neither the tool nor this PR has drifted. The structure supports that: every one of the eight deletions returns as a superset, the default policy still resolves to `v1.4`/`v1.3` as the final `else`, and the new behaviour is guarded by content-derived flags.

## The artifact reproduces, independently

This is the condition #8 attached to accepting Bonsai at all, so I ran it here rather than taking your log:

| check | result |
|---|---|
| source pack public, ungated, Apache-2.0, repo sha `abbae723028d` | confirmed |
| `Ternary-Bonsai-27B-Q2_0.gguf` MD5 `01493c0ade7d4fee416dfd874cdfcfd1` | exact, 7,165,121,600 bytes |
| repacked with this branch's `repack.py` | 851 tensors, 498 ternary slot-verified, 26.89 B ternary weights |
| output MD5 `8babb56d67b63f2308f883912a272bb3` | **exact** |

So the contract holds: a public Apache-2.0 source at a pinned revision reproduces the exact q27 Bonsai artifact byte-for-byte, on a machine that had never seen it.

I also confirmed your replay claim mechanically. The eight-file delta from the #21 merge `c82f6642` to `b91b98b4` is content-identical to the pre-merge delta, line for line, so everything above carries over and nothing needed re-reviewing.

The Metal Bonsai runtime path I have not exercised: the deterministic `Paris.` smoke with token-file MD5 `ddae10a2...` is yours, not mine. Neither of my Macs has the `gguf` module to repack locally, and I would rather not install packages on them than transfer 7.15 GB for it. Happy to if you want a second data point.

manthedan:
Thanks — both findings were correct. Addressed in `1b88a20e29d7212ccdd33d4bda2566ae64ba3ae3`.

**Finding 1:** `BONSAI_MODEL` is optional again. The engine-contract binary now accepts `OFFICIAL.q27 [BONSAI.q27]`; `test-metal-contracts` always runs the official engine/model/stream legs and supplies the Bonsai model only when configured, otherwise printing the Bonsai skip notice. The corrected binary builds cleanly with `-Wall -Wextra -Werror`, and a no-`BONSAI_MODEL` regression probe reaches the official engine-contract leg rather than failing at the Make prerequisite.

**Finding 2:** I removed the dspark/Q4_1 path rather than broadening this PR. No dspark, Q41, Q4_1_G32, metadata, or quant-policy path remains in the proposal. `repack.py` now adds only the T2/B1 Bonsai source formats named here.

I also reran the complete public-source repack from the pinned source on an independent Linux host in an isolated `gguf==0.19.0` environment: 851 tensors, all 498 ternary tensors slot-verified, 26.89 B ternary weights, 29.7% zeros, output size `7,154,196,992`, MD5 `8babb56d67b63f2308f883912a272bb3`.

The review-fix commit changes only `Makefile`, `src/metal/test_metal_engine_contracts.cpp`, and `tools/repack.py`; Metal runtime and shader sources are unchanged from the tip you exercised. I updated the PR body with the exact new tip/tree and separated exact-tip harness/repack evidence from the unchanged-runtime M4 evidence.

signalnine:
Both fixes verified at `1b88a20e`. Gated from my side.

**Finding 1.** Ran `test-metal-contracts` without `BONSAI_MODEL` on a base M4, which was a hard `exit 2` before:

```
Bonsai engine contracts: SKIP (set BONSAI_MODEL to enable)
Metal q4s engine contracts: OK
Metal constrained MTP, stale-snapshot, and chunk-logit contracts: PASS
Metal stream format: OK
```

Skip with a notice, official coverage intact. That is the shape I wanted.

**Finding 2.** `dspark`, `Q41` and `Q4_1` are at zero occurrences in `repack.py`. Removing the path rather than documenting it is the better of the two options I offered, and it keeps this PR to what its title says.

**Your unchanged-runtime claim holds mechanically.** `q27_kernels.metal` hashes to `0761aedd28f1afc88e52c88a04c718e38a6efbf4` at both `b91b98b4` and `1b88a20e`, so the canonical I measured on the earlier tip carries without a re-run, and the fix commit really is confined to the three files you named.

**`repack.py` still produces the official tier byte-identically.** Since the fix commit touched shared constant lines (`DTYPE_NAMES`, the `GROUP_*` tuple) even though the removed logic was Bonsai-only, I re-ran the check rather than reasoning from the diff: q4s repacked from `Qwen3.6-27B-MTP-BF16.gguf` with this tip's `repack.py` gives MD5 `7e5454e0c0ded717136ad3e42634ba25`, identical to the published `qwen36-27b-mtp-q4s.q27`.

## Summary of what I exercised on this PR

| check | result |
|---|---|
| public-source artifact reproduction, independent machine | source MD5 `01493c0a...` and output MD5 `8babb56d...` both exact |
| official q4s repack byte-identity, at three separate tips | identical to published, `cmp` clean |
| q4s Metal canonical with Bonsai kernels present | `f3010955...`, ALL PASS |
| `test-metal`, `test-metal-contracts`, `test-metal-validation` | pass |
| replay onto the #21 merge | eight-file delta content-identical, line for line |

The Metal Bonsai runtime path is the one thing I have not exercised. The deterministic `Paris.` smoke and token-file MD5 `ddae10a2...` are yours, now on three hosts including an isolated `gguf==0.19.0` Linux environment, which is better evidence than a second run from me would have been.

One note for the record rather than for action. Partway through verifying the repack I hit a single differing byte in a 15.46 GB output, one bit in `blk.33.ffn_down.weight`. It did not reproduce: the next run was byte-identical, and the machine has non-ECC memory with no EDAC. So it was a hardware bit flip on my side, not anything in this PR, and I mention it only because I nearly reported it as a regression. The lesson generalises to anyone repacking large artifacts: verify against `CHECKSUMS.md5` after every repack, because nothing else will tell you.

Thanks for taking both findings straight. Merging.

PR #26, opened 2026-08-21
vendor: cpp-httplib 0.18.3 -> 0.53.1

kaushikvira:
## Summary

Pure vendor bump of `third_party/httplib.h` from cpp-httplib 0.18.3 to 0.53.1. No source edits — the server API surface is unchanged across this range, so `src/` is untouched.

## Why

- **CVE-2025-66570** (security-relevant): header shadowing allowed an attacker-supplied header to shadow a parsed one, enabling auth bypass on the HTTP server. Fixed in the 0.48+ line.
- DoS fixes accumulated in 0.19..0.53 (header/body size handling, slowloris-class stalls).

## Scope

- `third_party/httplib.h` only: +18,594 / −7,023 lines (the vendored library itself)
- Zero changes to `src/`, `tools/`, `Makefile`, or tests

## Verification

Built and gated on this box (5090, CUDA 12.8, gcc):

- `build/q27` + `build/q27-server` compile clean
- `test_auth`, `test_auth_integration`, `test_openai_bridge`, `test_chat_completions_integration`, `test_tool_drift`, `test_loader_contracts` — all PASS

### DIFF
--- a/third_party/httplib.h
+++ b/third_party/httplib.h
@@ -1,19 +1,22 @@
 //
 //  httplib.h
 //
-//  Copyright (c) 2024 Yuji Hirose. All rights reserved.
+//  Copyright (c) 2026 Yuji Hirose. All rights reserved.
 //  MIT License
-//
-// q27 local patch: expose nonblocking peer-liveness callbacks to long-running
-// server handlers and content providers. Keep this patch when updating the
-// vendored cpp-httplib release; metal_server.cpp depends on both callbacks.
-//
 //
 
 #ifndef CPPHTTPLIB_HTTPLIB_H
 #define CPPHTTPLIB_HTTPLIB_H
 
-#define CPPHTTPLIB_VERSION "0.18.3"
+#define CPPHTTPLIB_VERSION "0.53.1"
+#define CPPHTTPLIB_VERSION_NUM "0x003501"
+
+#ifdef _WIN32
+#if defined(_WIN32_WINNT) && _WIN32_WINNT < 0x0A00
+#error                                                                         \
+    "cpp-httplib doesn't support Windows 8 or lower. Please use Windows 10 or later."
+#endif
+#endif
 
 /*
  * Configuration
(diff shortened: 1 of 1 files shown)


signalnine:
merged, thanks -- this is a good one to have.

i checked the vendored blob against upstream rather than reading the diff.
nobody reviews a 21k-line single header by eye, so i compared the bytes:

```
bc69d53636a8757cb24a1deb9880bf7e2fdae3a80bbc759e145b8c80913cbfa3  yhirose/cpp-httplib @ v0.53.1
bc69d53636a8757cb24a1deb9880bf7e2fdae3a80bbc759e145b8c80913cbfa3  third_party/httplib.h in this PR
```

identical, 747115 bytes both. one file touched, rebased on current master, and
test_auth / test_auth_integration / test_openai_bridge /
test_chat_completions_integration / test_tool_drift all pass built against it.

two corrections to the writeup, neither of which changes the merge.

CVE-2025-66570 is real (GHSA-xm2j-vfr9-mg9m) but it was fixed in 0.27.0, not the
0.48+ line. i went looking for it in the 0.48+ release notes, didn't find it,
and briefly thought it was made up, so it's worth pinning down.

and it doesn't reach us. q27 never reads REMOTE_ADDR / REMOTE_PORT /
LOCAL_ADDR / LOCAL_PORT, and uses neither X-Forwarded-For nor trusted_proxies;
auth keys on Authorization and x-api-key. so the header-shadowing bypass isn't a
live risk here. the bump still earns its place on the rest of the range: CRLF
injection in chunked trailers, the TLS chain verification bypass for IP-literal
hosts, the websocket use-after-free, and the chunked decoder that accepted
negative chunk sizes.

one thing still open on my side, not yours: src/metal/metal_server.cpp also
includes httplib and builds with -Werror on clang, and that's exactly how a
previous change broke the metal target while the CUDA side stayed green. both my
macs were asleep when i merged, so i'll confirm that build and follow up if it
needs anything.

PR #28, opened 2026-08-21
toolgram: XML dialect for --constrain-tools; enclosure result guard; tail recovery

kaushikvira:

Three related tool-call parsing fixes for Qwen3.8's XML dialect.

The metal `is_alive` -> `is_writable` rename this originally depended on has already landed as `c578411`, so this applies to current `master` on its own. It touches `src/metal/metal_server.cpp` only to give `ToolMaskCache` its template argument (item 3).

## 1. The hallucinated-result guard was vetoing REAL calls

The guard was a whole-string `find("<result>")` / `find("<output>")` gating the entire mode 17/20/21/22 recovery block, so one quoted `<output>` anywhere in a segment -- prose, a fenced transcript, a prior step's result -- skipped every recovery below it and a fully-formed mode-21 call leaked out as plain text:

```
Here is what the build printed:
<output>make: all targets up to date</output>

Now I'll fix the import.
<parameter=filePath>
/code/fileaction.go
</parameter>
<parameter=newString>
import "bytes"
</parameter>
</function>
```

Presence cannot discriminate: this and the genuine hallucinated-result shape BOTH contain a complete `<output>...</output>` block before the call span. The test is structural -- does the result element CONTAIN the candidate call?

```
(a) a <result>/<output> tag INSIDE the span      -> invented output posing as a
                                                    parameter value; refuse
(b) the span begins inside an UNCLOSED such tag  -> the call is part of the
                                                    invented result block; refuse
otherwise (every block closed before the span)   -> prior context; ACCEPT
```

(b) is what catches the real shapes: the model opens `<result>` and never closes it, so the `<tool_name>` / `<parameter=` span that follows is enclosed by it. The two cases differ by exactly one closing tag. Depth counting, not a last-open scan, so closed-then-reopened still reads as enclosed.

Extracted as `hallucinated_result_around()` and applied per-mode against each mode's OWN span (16, 17a, 17b, 20, 21, 22) instead of one coarse veto over all of them. Mode 16 carried a second copy of the whole-string guard with the same latent bug and is on the shared helper now.

**Note for review:** scoping only the END of the search window does NOT fix this -- a closed block before the call is still inside the window and still vetoes it. And dropping the coarse guard while replacing it with a mode-21-only check strips modes 20/22 of protection, trading a dropped call for an EXECUTED hallucination. Both are pinned by tests.

## 2. Unclosed-`<tool_call>` tail recovery

bb60661 routed a CLOSED wrapped TOOL segment through the drift chain but left the unclosed one as raw text. `recover_unclosed_tool_tail()` now routes that tail too, with `allow_eof_repair=false`: an inner call whose JSON/XML is COMPLETE and merely lost its `</tool_call>` to truncation recovers, while a value cut mid-parse stays text. Executing half a Write is still refused. Wired at all six handler sites (chat, chat-stream, messages, messages-stream, responses, responses-stream).

Defence-in-depth: this and `resolve_ordered_tool_segments` now refuse a call whose `source_begin` is npos or overlaps the emitted cursor, because `substr(cursor, begin - cursor)` would underflow to `SIZE_MAX` and throw inside the request handler. Mode 20 stamps per-call spans rather than only first/last, which is what left middle calls at npos.

## 3. XML-dialect toolgram

`--constrain-tools` only spoke the 3.6 JSON body format, so on an XML-trained 3.8 the grammar fought the model's own trained emission. `ToolGrammarXml` constrains the body to `<function=NAME><parameter=KEY>VALUE</parameter></function>` instead, schema-aware: the parameter-key allowlist switches to that tool's declared parameters once the name completes. `<` inside a value is legal ONLY as the start of `</parameter>` or `</function>`, so a model reaching for `<result>` or a nested `<function=` hits a dead state and the decoder masks it out -- the JSON/XML chimera is prevented at the source rather than refused after the fact.

`ToolMaskCache` is templated on the grammar so both dialects share the caching and the allowlist-in-cache-key isolation (R1) unchanged. Metal stays JSON-only and warns once when it would matter instead of silently no-opping.

## Testing

`test_tool_drift` 124 -> 151 assertions, new ones red before the change; `test_toolconstrain` covers the XML grammar, its cache keying and dialect dispatch. Full `make test-tools` green under gcc.

**NOT VERIFIED:** no live-generation A/B (the survey harness needs a GPU host this work was not done on), and neither backend was compiled -- no CUDA and no Apple-silicon machine was available. The changes outside `src/metal/` are header-only C++ exercised by the CPU suites, but `build/q27-server` and `q27-metal-server` both want a real compile before anyone leans on this.

BUILDLOG entry included, per the log's convention for parser changes.

### DIFF
--- a/docs/BUILDLOG.md
+++ b/docs/BUILDLOG.md
@@ -14025,3 +14025,166 @@ both channels can reach it.
 
 Both backends green: `make test-tools` under gcc on haight and clang on the M4,
 `q27-metal-server` links clean under `-Werror`.
+
+## 2026-08-21: the hallucinated-result guard was vetoing REAL calls -- presence is not enclosure
+
+Follow-on to (d). The report was the same sentence as always -- "sometimes I
+still see `<parameter=` / `</function>` in the output" -- but this time the
+bytes recovered cleanly when replayed in isolation and died in the session. The
+difference was the CONTEXT they arrived in, not the call.
+
+**The shape.** A perfectly recoverable mode-21 openerless call, in a tail that
+also quoted an earlier step's output:
+
+```
+Here is what the build printed:
+<output>make: all targets up to date</output>
+
+Now I'll fix the import.
+<parameter=filePath>
+/code/fileaction.go
+</parameter>
+<parameter=newString>
+import "bytes"
+</parameter>
+</function>
+```
+
+The hallucinated-result guard was a **whole-string** `find("<result>")` /
+`find("<output>")` gating the ENTIRE mode 17/20/21/22 block. One quoted
+`<output>` anywhere in the segment -- prose, a fenced transcript, a prior
+step's result -- and every recovery below it was skipped. The call was sitting
+right there, fully formed, and the guard threw the chain away.
+
+Worth recording precisely because 2026-08-20 (b) **ruled this guard out as a
+hypothesis**: "the hallucinated-result guard (no `<result>`/`<output>` in any
+capture)". That was true of those captures and false as a general claim. The
+instrument was a corpus of failures that had already been collected; a failure
+mode whose trigger is *the presence of a tag none of the captures contained*
+cannot appear in it. Third variant of the same lesson from (b), (c) and (d):
+**absence of evidence in a catalogue built from past failures is not evidence
+of absence.**
+
+### The obvious repair does not work, and it is worth saying why
+
+The first instinct is "scope the guard to the call span", and the first
+implementation of that scopes the wrong end: search for a complete
+`<output>...</output>` from **position 0** up to `span_end`. That bounds the END
+of the window and leaves the START at the beginning of the segment, so a closed
+block *before* the call is still inside the window and still vetoes it --
+precisely the bug, unchanged. Worse, dropping the coarse outer guard while
+replacing it with a span check that only covers mode 21 leaves modes 20 and 22
+(and the wrapped variant) with no protection at all, so three previously-pinned
+negatives start promoting invented output to calls.
+
+Recorded because the half-fix is more dangerous than the bug: it trades a
+dropped call for an EXECUTED hallucination, and it looks like progress.
+
+### The rule: enclosure, not presence
+
+Lexical presence cannot separate the two cases, because both have a complete
+`<output>...</output>` block before the call span. The discriminator is
+structural -- does the result element *contain* the candidate call?
+
+```
+(a) a <result>/<output> tag INSIDE the span      -> invented output posing as a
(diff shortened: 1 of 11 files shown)


kaushikvira:
I am following up on with this https://github.com/kaushikvira/q27/issues/2  this is issue in both xml/json schema

kaushikvira:
## ✅ Build + tests passed

Tested the exact PR head `4c18a342bc7df4a73195d3c0ea471e93849e70b2` on a local checkout (g++ -O2 -std=c++17, RTX 5090 host).

| Suite | Result |
|-------|--------|
| test_openai_bridge | PASS |
| test_think_resolve | PASS |
| test_stream_split | PASS |

signalnine:
holding this one in favour of #33, which carries this commit verbatim (i checked
-- identical trees) plus the two toolgram commits. no point merging the same
change twice.

the content is verified and not the blocker. details are on #33 so they're in one
place: the enclosure guard still refuses all four invented-result shapes with
only the false veto flipping, and i reproduced your span-underflow crash on
master before your PR existed. it builds green on all three targets too --
CPU suites on linux and macos, q27-server complete under nvcc, q27-metal-server
clean under -Werror once merged onto current master.

so this lands as soon as #33 is rebased. leaving it open until then rather than
closing, in case you'd rather split it back out.

kaushikvira:
Marking this closed-out per your note: #33 is rebased onto current master (`08c1680`) and flipped to **ready for review** — head `9fd3dc4`, its own 3 commits.

The rebase conflict in `src/api_common.h` was the one real merge: your batch rewrite of the mode-16 opener path (the "recover every call in a wrapped batch" follow-up) collided with this PR's switch of that mode onto `hallucinated_result_around`. Resolved as both — the batched loop now applies the per-span hallucination guard to each candidate span, so this PR's guarantee survives on top of the batching. `docs/BUILDLOG.md` and `tools/test_tool_drift.cpp` were append-order as you said.

`make test-tools` green from a clean-ish build on the rebased head (exit 0, all 8 suites pass, full table on #33). Still no CUDA / Apple-silicon host here, so the two server binaries still want their real compiles — unchanged from before.

Will leave this open per your call, to be closed when #33 merges — say if you'd rather split it back out and I'll do it from the rebased tree.

kaushikvira:
Closing per the handoff: this change is fully carried by #33 (identical subtree, now rebased and ready at `9fd3dc4`). Merging #28 separately would apply the same commits twice. If we ever need it split back out, the 3 commits live on `kv/toolgram-required-args` and can be re-opened from there.

PR #29, opened 2026-08-21
metal: follow cpp-httplib 0.53.1 DataSink rename is_alive -> is_writable

kaushikvira:
a317bc8 vendored cpp-httplib 0.18.3 -> 0.53.1, which renames DataSink::is_alive to DataSink::is_writable (third_party/httplib.h:1432). src/metal/metal_server.cpp still calls sink.is_alive() at four sites, so q27-metal-server does not compile on the current tip.

Renames the four call sites. No behaviour change -- 0.53.1 wires is_writable to the same strm.is_peer_alive() the old field carried.

Worth noting why this survived the vendor bump: every CPU suite and the CUDA build are blind to src/metal/, so 'make test-tools' is fully green on a tree where one of the two backends cannot build at all. A per-backend compile gate would have caught it.

NOT VERIFIED: no Apple-silicon host was available, so this was not compiled. The break is a source-level fact about the vendored header's API (the field does not exist under the old name), and the replacement is the field 0.53.1 defines, but it wants a real 'make metal' before anyone leans on it.

### DIFF
--- a/src/metal/metal_server.cpp
+++ b/src/metal/metal_server.cpp
@@ -2512,7 +2512,7 @@ int main(int argc,char** argv) {
                         auto outcome=runtime.guard_engine([&]{
                             return runtime.run(ids,n,sampling,stops,emit,
                                 std::vector<std::string>{},snap_hint,id,nullptr,false,
-                                [&]{ return sink.is_alive(); });
+                                [&]{ return sink.is_writable(); });
                         });
                         if(outcome.finish==Runtime::Finish::Cancelled) { sink.done(); return false; }
                         // Terminal chunk with a real finish_reason before [DONE]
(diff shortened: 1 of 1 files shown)


signalnine:
merged, thanks -- and this one is on me. i merged the vendor bump having written
"NOT VERIFIED: the metal build" into the merge commit, which is a fairly
embarrassing way to be right.

checked the rename independently before taking it: httplib.h:1432 declares
`std::function<bool()> is_writable` wired to `strm.is_peer_alive()` at three
sites, DataSink has no `is_alive` member under 0.53.1, and there are exactly
four call sites, all in metal_server.cpp. src/server.cu never called it, so the
CUDA build was never affected.

still not compiled, same reason as before -- no apple-silicon host reachable
when i merged. so this is a source-level fix to a source-level break, and
q27-metal-server has yet to be built on either side of it. i'll confirm as soon
as i can get on a mac.

the wider point is the one i'm actually taking away: make test-tools and the
CUDA build are both blind to src/metal/, so the suite is fully green on a tree
where one of the two backends cannot compile at all. that's the gap, not the
missed rename. a per-backend compile gate would have caught it at the bump.

PR #30, opened 2026-08-22
stream_split: one unbalanced quote disarmed tool calling for the rest…

kaushikvira:
… of the turn

JsonStringLexState never closed a string on a raw newline, so a single unbalanced double quote in prose left in_string set for every byte that followed. display_text_context_is_executable() then reported "not executable" for the rest of the generation, the next genuine <tool_call> was classified non-structural, and stream_split.h's guard emitted it as VISIBLE TEXT -- the call was lost AND the raw marker leaked to the user:

    I ran echo "---
    now calling.
    <tool_call>{"a":1}</tool_call>      <- marker shown as text, no call

A sentence quoting a shell command was enough. RFC 8259 requires control characters inside a JSON string to be escaped, so "\n" is two bytes and never one: an unescaped line break proves the quote was prose, and the string ends there.

SCOPED TO DISPLAY CLASSIFICATION, deliberately. The parser shares this lexer, and its repair paths synthesize newlines -- mode 10 reconstructs a dropped opener as "...;\n{\"name\": ...". Applying the rule unconditionally made the drift chain rescue a second same-line call that test_openai_bridge refuses on purpose, because an invalid call followed by a valid one on one line has an ambiguous boundary. Widening a rescue boundary is how you execute a call the model did not ask for, so the rule is opt-in via
JsonStringLexState::display(), used only by StreamSplitter's text_string_state. "Is this marker structural?" and "where does a rescued call begin?" are different questions and no longer share an answer.

Does not reopen the fence-skip hole: a marker quoted inside a fenced block stays visible text, unbalanced quote or not. Pinned.

KNOWN LIMIT, now pinned so a change is deliberate: an UNCLOSED fence still runs to the end of the generation, so a real call after it is refused. Conservative direction is right -- never execute an echoed marker -- but it costs the turn.

4 new assertions in test_stream_split (byte-at-a-time and chunked). Full make test-tools green from a clean build.

NOT VERIFIED: not compiled for either backend -- no CUDA and no Apple-silicon host was available. Header-only change exercised by the CPU suites.

### DIFF
--- a/src/markdown_lex.h
+++ b/src/markdown_lex.h
@@ -9,6 +9,19 @@ namespace q27 {
 struct JsonStringLexState {
     bool in_string=false;
     bool escaped=false;
+    // Opt-in, DISPLAY-CLASSIFICATION ONLY (see consume()). Off by default
+    // because the drift chain synthesizes newlines while repairing dropped
+    // openers, and closing a string there would widen which repairs the
+    // parser accepts -- a rescue-boundary decision, not a display one.
+    bool close_string_at_newline=false;
+
+    // State for asking "is a marker in VISIBLE text structural?", where an
+    // unterminated quote must not disarm the rest of the turn.
+    static JsonStringLexState display() {
+        JsonStringLexState s;
+        s.close_string_at_newline=true;
+        return s;
+    }
     // Active containers use their opener. A top-level undecided opener uses
     // 'o'/'a' until the first non-space byte proves it is JSON-like.
     std::string containers;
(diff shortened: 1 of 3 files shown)


signalnine:
merged, thanks. good catch -- a stray double quote in prose disarming tool
calling for the rest of the turn is the kind of thing that would have read as
"the model just stopped calling tools sometimes".

the RFC 8259 argument holds: a raw newline can't appear inside a JSON string, so
the quote was prose and the string ends at the break.

what made this mergeable rather than merely correct is the scoping. closing
strings sooner means in_string is false more often, so MORE markers read as
structural, and that's the direction where you start executing echoed markers.
keeping it opt-in via display(), on the splitter's state only, means the rescue
boundary question is untouched -- and your note about mode 10 synthesizing
newlines during repair is exactly why. i'd have been tempted to just fix the
lexer.

fence negatives and the same-line case are pinned, and i like that the unclosed
fence limit is a test rather than a comment, so changing it has to be deliberate.

verified here: full make test-tools green in a clean worktree.

PR #31, opened 2026-08-22
stream_split: an in-value </tool_call> is provisional, not the closer

kaushikvira:

The XML dialect has no escape mechanism, so a tool call whose parameter value legitimately contains the protocol's own bytes — a doc, a verifier, a bug write-up — collided with the channel delimiter and was destroyed.

**The document describing this bug was itself truncated at 651 bytes by this bug**, mid-sentence, exactly where it was about to type the closing tag. That is the whole report.

## The failure

````
<tool_call>
<function=write>
<parameter=content>
The model emitted:
```
<tool_call>
{"name":"bash"}
</tool_call>          <-- StreamSplitter closed the TOOL channel HERE
```
and never closed it.
</parameter>
</function>
</tool_call>
````

`StreamSplitter`'s TOOL branch was a bare `hold.find("</tool_call>")`. The TEXT channel has careful structural-vs-inert logic — that is what stops a fenced example from firing a call — but the TOOL channel had none at all. So the call was cut at the inner tag, failed to parse, recovered **nothing** (`calls=0`), and the trailing `</parameter></function>` leaked as visible text.

Reproduced deterministically on `master`. The model's own diagnosis in the live session was correct: *"the harness is interpreting it as its own delimiter."*

## Why the obvious fix is wrong

"Ignore `</tool_call>` inside a parameter value" breaks a case that works today. Models routinely **forget** the final `</parameter>` and close with `</tool_call>`; `parse_native_xml_call` tolerates exactly that (the 2026-08-15 EOF leniency). Ignoring the tag there means the call never closes and is lost in the other direction.

The two cases are indistinguishable *at the moment the tag arrives*, and become distinguishable later:

- if the value **closes** afterwards → the tag was content
- if the generation **ends** first → the tag was the closer

## The rule: provisional closers

An in-value `</tool_call>` is **provisional**. Bytes from it are **held rather than emitted**, so the decision stays open until the evidence arrives. A subsequent `</parameter>` or `</function>` proves it was content and releases the held bytes; `flush()` reaching end-of-generation with one still pending promotes it to the real closer and splits exactly where the old code did. Both cases work, and nothing is guessed at the point of ambiguity.

## The streaming half, which nearly shipped broken

The value-state machine also has to survive `feed()` boundaries. The TOOL channel previously held back only a partial `</tool_call>`, so a `<parameter=` split across two tokens was never recognised, `tool_in_value` stayed false, and the next in-value `</tool_call>` read as structural.

My first version therefore passed on whole-string input and **failed at chunk sizes 1 and 5** — content truncated, closers still leaking. That would have been a fix that works in tests and does nothing in production, since streaming is the only mode that matters.

`tool_keep()` now holds back a partial prefix of any of the four dialect markers, plus an unterminated `<parameter=` opener whose `>` has not arrived. **Every fixture is pinned at chunk 1, 5 and whole-string**, because whole-string-only would have shipped this broken.

## Testing

7 new assertions in `test_stream_split`: content-with-closers recovers with the value byte-identical (chunked and bytewise), an unterminated value still promotes its provisional to the closer, trailing bytes after a promoted closer route to TEXT, a split `<parameter=` opener keeps its state, and a zero-parameter call closes immediately. Full `make test-tools` green from a clean build.

## Notes

- Independent of the other open PRs; no conflicts.
- **Known limit:** this reasons about the XML dialect's value spans. A JSON-dialect body whose string value contains `</tool_call>` has the same collision and is **not** covered.
- **Review focus:** this changes closer detection on every request, which is the riskiest surface in the splitter. Worth attention on large `content` values and under batching.
- **Not verified:** neither backend compiled — no CUDA and no Apple-silicon host available.

### DIFF
--- a/docs/BUILDLOG.md
+++ b/docs/BUILDLOG.md
@@ -14025,3 +14025,84 @@ both channels can reach it.
 
 Both backends green: `make test-tools` under gcc on haight and clang on the M4,
 `q27-metal-server` links clean under `-Werror`.
+
+## 2026-08-22: the dialect has no escaping, so writing ABOUT a tool call destroyed the tool call
+
+The document describing this bug was truncated at 651 bytes BY this bug, mid
+sentence, exactly where it was about to type `</tool_call>`. That is the whole
+report.
+
+The XML dialect provides no escape mechanism, so a call whose parameter value
+legitimately contains the protocol's own bytes -- a doc, a verifier, a bug
+write-up -- collides with the channel delimiter:
+
+```
+<tool_call>
+<function=write>
+<parameter=content>
+The model emitted:
+```
+<tool_call>
+{"name":"bash"}
+</tool_call>          <-- StreamSplitter closed the TOOL channel HERE
+```
+and never closed it.
+</parameter>
+</function>
+</tool_call>
+```
+
+`StreamSplitter`'s TOOL branch was a bare `hold.find("</tool_call>")`. The TEXT
+channel has careful structural-vs-inert logic (that is what keeps a fenced
+example from firing a call); the TOOL channel had none at all. So the call was
+cut at the inner tag, failed to parse, recovered NOTHING, and the trailing
+`</parameter></function>` leaked as visible text -- `calls=0`. Measured on
+master and on every branch: the model's own diagnosis in the live session was
+right, "the harness is interpreting it as its own delimiter".
+
+### Why the obvious fix is wrong
+
+"Ignore `</tool_call>` inside a parameter value" breaks a case that works
+today. Models routinely FORGET the final `</parameter>` and close with
+`</tool_call>`; `parse_native_xml_call` tolerates exactly that (the 2026-08-15
+EOF leniency). Ignoring the tag there means never closing, and the call is lost
+in the other direction.
+
+The two cases are indistinguishable at the moment the tag arrives, and become
+distinguishable later: if the value CLOSES afterwards, the tag was content; if
+the generation ENDS first, the tag was the closer.
+
+### The rule: provisional closers
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by kaushikvira** [comment]
## ✅ Build + tests passed

Tested the exact PR head `d65d82fa1d2147e591cfe680710ec21c254c5809` on a local checkout (g++ -O2 -std=c++17, RTX 5090 host).

| Suite | Result |
|-------|--------|
| test_openai_bridge | PASS |
| test_think_resolve | PASS |
| test_stream_split | PASS |

**Comment by signalnine** [comment]
merged, thanks. the provisional-closer design is the right call and i want to
say why rather than just taking it: "ignore `</tool_call>` inside a value" was
where i'd have gone, and your point that models routinely drop the final
`</parameter>` and close with `</tool_call>` -- which `parse_native_xml_call`
tolerates on purpose -- is exactly the case that would have broken. holding the
bytes until the evidence arrives is better than picking a side at the ambiguous
moment.

since this changes closer detection on every request, i checked it against
master independently of your tests, at several chunk sizes:

```
case                     master      this branch
value contains closer    LOST        recovered     (chunk 0/1/5/17)
EOF promotes closer      ok          ok            (chunk 0/1/5/17)
ordinary call            ok          ok            (chunk 0/1/5/17)
```

your note about the first version passing whole-string and failing at chunk 1
and 5 is the most useful thing in the PR. a fix that works in tests and does
nothing in production is worse than no fix, and pinning every fixture at 1, 5
and whole-string is the right response.

built both backends: q27-server complete under nvcc, q27-metal-server clean
under -Werror on apple clang 21, CPU suites green on linux and macos.

noted the JSON-dialect limit. same collision, not covered, worth a follow-up.

PR #32, opened 2026-08-22
parser: a repeated <tool_call> opener must not become the user's answer

kaushikvira:

A repeated `<tool_call>` opener became the user's entire visible answer for a turn.

## The live failure

Session `01a026cd` (`rid chatcmpl-q27-8`): after 8 clean tool calls the model emitted the opener **twice** and hit EOS at 14 tokens. The whole visible assistant message was the literal string `<tool_call>`, with no tool call and `stopReason: "stop"`.

```
<tool_call>     <- structural: the splitter erases it, enters the TOOL channel
<tool_call>     <- inside TOOL only "</tool_call>" delimits, so this is BODY CONTENT
(EOS)
```

## Why it reaches the user

- `unfinished_tool_wrapper()` reports **false** here — it detects length/budget truncation, not EOS-inside-wrapper — so `take_unclosed_final_tool_segment()` and the tail-recovery path never run.
- `emit_tool()` then fails to parse, the drift chain finds nothing, and the fallback prints `ToolCall::raw`.
- `tc.raw = seg` (`api_common.h`) is the body verbatim, and the body **is** the literal string `<tool_call>`.

So the marker was never leaked by the splitter — it arrived as body content. A single opener is handled correctly; two is the trigger. Three leaks two of them.

## The fix

`only_dialect_control_bytes()`, applied at the two fallback points — `resolve_ordered_tool_segments()`'s TOOL branch and the streaming `emit_tool()` in both `server.cu` handlers — with a stderr line so the lost turn is not silent.

**Narrow by construction:** whitespace plus `<tool_call>`/`</tool_call>` only. Tags that can carry a payload (`<function=...>`, `<parameter=...>`) are **not** stripped, because a body containing those is a malformed *call* whose bytes the user may need to see. Drift stays visible; only provably content-free residue is suppressed. Pinned in both directions — a malformed body still reaches the user, a well-formed call is unaffected.

## Deliberately not done

This does **not** widen `unfinished_tool_wrapper()` to treat "channel is TOOL at EOS" as unfinished. That is arguably the more general fix, but finish detection gates every request, and the currently-*working* case (unclosed wrapper, complete JSON, EOS) already recovers through `emit_tool()`. Rerouting it through the no-EOF-repair tail path risks regressing it. Recorded as an open question rather than changed blind.

## Scope

The turn is still lost — a model that emits two openers and stops has produced nothing executable. This decides only what the user sees, and that it gets logged.

## Testing

13 new assertions in `test_tool_drift` (the repeated-opener shapes, the predicate's boundaries in both directions, and a malformed body that must still be visible). Full `make test-tools` green from a clean build.

## Notes

- Independent of the other open PRs; no conflicts.
- **Not verified:** neither backend compiled — no CUDA and no Apple-silicon host available.

### DIFF
--- a/src/api_common.h
+++ b/src/api_common.h
@@ -4322,6 +4322,42 @@ struct OrderedToolOutput {
     }
 };
 
+// A TOOL body that recovered NOTHING and consists only of dialect control
+// bytes carries no information for the user -- showing it just leaks protocol.
+//
+// The shape that made this necessary (2026-08-22, live): the model emitted the
+// opener TWICE and then hit EOS --
+//
+//   <tool_call>          <- structural; the splitter erases it, enters TOOL
+//   <tool_call>          <- inside TOOL only "</tool_call>" delimits, so this
+//   (EOS)                   is BODY CONTENT, not a marker
+//
+// unfinished_tool_wrapper() reports false (it detects length/budget truncation,
+// not EOS-inside-wrapper), so the unclosed-tail path never runs. emit_tool()
+// then fails to parse, the drift chain finds nothing, and the fallback prints
+// ToolCall::raw -- which IS the literal string "<tool_call>". The user's
+// visible answer for the whole turn was that one marker.
+//
+// Deliberately NARROW: whitespace plus <tool_call>/</tool_call> only. Dialect
+// tags that can carry a payload (<function=...>, <parameter=...>) are NOT
+// stripped, because a body containing those is a malformed CALL whose bytes the
+// user may need to see -- that is drift, and drift stays visible. This only
+// suppresses residue that is provably content-free.
+//
+// The turn is lost either way; this decides what the user sees, and callers log
+// so the loss is not silent.
+inline bool only_dialect_control_bytes(const std::string& s) {
+    size_t i=0;
+    bool saw_marker=false;
+    while(i<s.size()) {
+        if(isspace((unsigned char)s[i])) { i++; continue; }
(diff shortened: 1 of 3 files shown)


kaushikvira:
## ✅ Build + tests passed

Tested the exact PR head `a7597dfc0306e5a6edf9681d33a5a6c4285a01af` on a local checkout (g++ -O2 -std=c++17, RTX 5090 host).

| Suite | Result |
|-------|--------|
| test_openai_bridge | PASS |
| test_think_resolve | PASS |
| test_stream_split | PASS |

signalnine:
merged, thanks. the diagnosis is exactly right and i verified it against master:
doubled opener at EOS gives `visible="<tool_call>"` on master and `visible=""`
here, calls=0 either way.

the turn is lost regardless, so this is only about what the user sees -- but a
turn whose entire visible answer is the literal string `<tool_call>` is worse
than an empty one, and it now logs instead of vanishing.

i particularly agree with what you did NOT do. widening
`unfinished_tool_wrapper()` to treat "channel is TOOL at EOS" as unfinished is
the more general fix, but finish detection gates every request and the case that
currently works (unclosed wrapper, complete JSON, EOS) already recovers through
`emit_tool()`. rerouting that through the no-EOF-repair tail path to fix a
cosmetic leak would be a bad trade. recording it as an open question is the
right resolution.

the narrowness is what makes it safe: stripping only whitespace and the bare
openers/closers, leaving `<function=`/`<parameter=` bodies visible, keeps
genuine drift in front of whoever has to debug it.

built both backends, CPU suites green on linux and macos.

PR #33, opened 2026-08-22
toolgram: enforce required args and reject duplicate keys under --constrain-tools

### LINKED ISSUE #34: --constrain-tools: toolgram enforces XML shape but not schema-valid args (required params dropped, duplicates allowed)
Relocated from [kaushikvira/q27#2](https://github.com/kaushikvira/q27/issues/2) — the fix that addresses this is [PR #33](https://github.com/signalnine/q27/pull/33) (currently rebased and ready; it depends on the XML-dialect work already carried by that PR).

## Summary

`--constrain-tools` (toolgram / `ToolGrammarXml`) guarantees tool calls are **well-formed XML** but does **not** guarantee the arguments are **schema-valid**. The API schema's `required` array is dropped before the grammar is built, the grammar has no per-call completion check, and duplicate parameter keys are accepted. Result: the server emits `tool_calls` with missing required arguments (or even `{}`) while returning 200 and logging zero drift — the agent's client rejects them and the session dead-loops. Observed live with a real agentic session against `qwen38-27b-mtp` on the q27 profile.

## Observed failure (live, one session)

A coding agent was asked to write a file (`verify_tool_calls.py`). The `write` tool requires `path` + `content`. Every attempt came back schema-invalid:

```
Validation failed for tool "write":
  - path: must have required properties path, content
  Received arguments: {}
...
Validation failed for tool "write":
  - path: must have required properties path
  Received arguments: {"content": "..."}
```

The server log confirms the emitted XML contained **only** `<parameter=content>` — `<parameter=path>` was never produced, and the grammar accepted `</function></tool_call>` anyway:

```
[toolgram] engaged (rem=0, dialect=xml)
... <function=write>
<parameter=content>
#!/usr/bin/env python3
...
</parameter>
</function>
</tool_call>
[toolgram] call closed
```

The agent retried (bash heredocs, "write in parts") and kept failing; the file was never written (0 bytes).

A second failure mode compounded it: the file being written **itself contained the literal text `</parameter> <parameter=content>`** (it was a tool-call verifier), and the model emitted those as real tags mid-content, producing a **second `<parameter=content>`** — the grammar accepted the duplicate key and the call closed with mangled, doubled content.

## Why the existing safety nets caught nothing

- `[drift] UN-RESCUED` / `Q27_DRIFT_CORPUS` only fire when the **fallback parser** fails to parse. These calls parsed cleanly (well-formed XML), so the drift corpus stayed empty and the log showed all-200 — the system looked healthy while the session was actually broken.
- The config's claim that `--constrain-tools` is "a hard correctness guarantee" is only true for shape/syntax, not for argument completeness.

## Root cause

1. `tool_param_keys_per_name()` (`src/api_common.h:2058` and the `json` overload at `:2080`) extracts only the parameter **key allowlist** from `properties`. The schema's `required` array is parsed nowhere and never reaches the grammar.
2. `ToolGrammarXml` (`src/toolgram.h:478+`) tracks which parameter keys are *allowed* (`KEY` state, `:631-:649`), but has no notion of *which keys were emitted*. In the `DONE_`/`CT_CLOSE` states (`:676-:697`) it accepts `</function>` → `</tool_call>` with **any subset of params — including zero** — and the `KEY`→`GT2` transitions allow the same key to open again (no duplicate rejection).
3. The JSON-dialect grammar `ToolGrammar` (`src/toolgram.h:20+`) is **not schema-aware at all**: its `reset()` (`:22`) takes only `tool_names`, and the `arguments` object is constrained purely as generic valid JSON (`J_VALUE`/`J_KEY`/number/string states). No param-key allowlist, no required tracking — so it would emit `{"content": "..."}` (missing `path`) and even undeclared keys just as freely.
4. The **only** consumer of `required` anywhere — `src/api_common.h:2878-:2879` inside `infer_tool_name()` (`:2845`) — is the drift-mode-6 *post-hoc* rescue that guesses which tool an orphaned args object belongs to. It's after-the-fact **selection**, not decode-time **enforcement**, and can't help here: the call was well-formed XML that already named `write`.

So `--constrain-tools` prevents illegal characters and nested tags, but cannot prevent the two most common failure classes of free-generation tool calls: **missing required arguments** and **duplicate/misordered arguments** — in **both** dialects.

## Code references (for the fix)

- `src/api_common.h:2058` — `tool_param_keys_per_name(const OpenAIToolSelection&)` (XML allowlist builder; drops `required`)
- `src/api_common.h:2080` — `tool_param_keys_per_name(const json&)` (same, array overload)
- `src/server.cu:2065,2285,2678` — callers that build `params_per_name` and call `tc.begin(...)`
- `src/toolconstrain.h:41-68,232` — `params_per_name` storage + `tg_xml.reset(names, params_per_name)`
- `src/toolgram.h:478-706` — `ToolGrammarXml` (allowlist at `:631-:649`; close-without-required at `DONE_` `:676`; no duplicate rejection at `KEY`/`GT2`)
- `src/toolgram.h:20-420` — `ToolGrammar` (JSON dialect; no schema params at all)
- `src/api_common.h:2845-2905` — `infer_tool_name()` (only `required` consumer; post-hoc selection, not enforcement)

## Impact

Makes q27 effectively unusable for agentic traffic (Claude-Code style loops): the most common operation (writing files) can fail with zero signal in the server log, speculative/decode stats, or the drift corpus. The failure is only visible at the client's schema-validation layer.

## Expected / suggested fix

- Thread each tool's `required` array (from `parameters.required`) into `ToolGrammarXml` together with `params_per_name`.
- Track emitted keys per call; **mask** `</function>` (i.e. refuse `DONE_`) until every required key has appeared; **mask** duplicate key opens.
- (Observability, optional) record schema-invalid emissions (missing required / duplicate keys) even when the XML parses, so this class is visible in the log/corpus instead of silent.

Worst case after the fix: an empty-but-present `path` (schema-valid) instead of `{}` / missing-`path` / doubled-`content` garbage — client sees a valid call and the agent can proceed.

## Environment

- Engine: `~/work/q27`, commit `da3d2fc` (`parser: a repeated <tool_call> opener must not become the user's answer`)
- Server command: `q27-server qwen38-27b-mtp.q27 ... --port 8000 --constrain-tools`
- Model: `qwen38-27b-mtp` (`.q27` + `.tok`)
- Profile: kv's `q27-5090` serving profile (`--constrain-tools`, XML dialect, batch on)

kaushikvira:
> Carries the #28 tree (XML-dialect toolgram / enclosure guard / tail recovery) plus the required-args + duplicate-key enforcement. #28 was closed in its favour; this PR merged via upstream `master`.

Fixes [signalnine/q27#34](https://github.com/signalnine/q27/issues/34) (relocated from [kaushikvira/q27#2](https://github.com/kaushikvira/q27/issues/2)).

`--constrain-tools` guaranteed that a tool call was **well-formed XML** and nothing else. It did not guarantee the arguments were **schema-valid**, so the server emitted calls with required arguments missing, returned 200, and logged no drift. The only component that noticed was the client's schema validator — on the far side of the wire.

## The live failure

The agent was asked to write a file. `write` requires `path` + `content`. Every attempt came back from the client:

```
Validation failed for tool "write":
  - path: must have required properties path, content
  Received arguments: {}
...
  - path: must have required properties path
  Received arguments: {"content": "..."}
```

and the server log for the same turn looked healthy:

```
[toolgram] engaged (rem=0, dialect=xml)
<function=write>
<parameter=content>
...
</parameter>
</function>
</tool_call>
[toolgram] call closed
```

`<parameter=path>` was never emitted and the grammar **accepted the close anyway**. The agent retried with heredocs and "write in parts"; the file stayed 0 bytes.

## Root cause

`ToolGrammarXml` was schema-aware for the **allowlist** only. `tool_param_keys_per_name()` extracts `properties`, so the grammar knew which keys *may* appear. `parameters.required` was parsed **nowhere** — the only consumer of `required` in the tree was `infer_tool_name()`, the post-hoc drift-mode-6 rescue that guesses which tool an orphaned arguments object belongs to. That is after-the-fact *selection*, not decode-time *enforcement*.

So the grammar had no notion of which keys *had* appeared: `DONE_` accepted `</function>` on any subset including the empty one, and `KEY` → `GT2` let the same key open twice.

Two failure classes fell straight through a constraint advertised as a hard correctness guarantee: **missing required arguments** and **duplicate keys**. The duplicate case compounded live, because the file being written was itself a tool-call verifier — its content quoted `</parameter>` and `<parameter=content>`, the model emitted those as real tags mid-value, and the grammar accepted the second `content` key, doubling the value.

## Why nothing caught it

`[drift] UN-RESCUED` and `Q27_DRIFT_CORPUS` fire when the **fallback parser fails**. These calls parsed perfectly — well-formed XML naming a declared tool. So the corpus stayed empty, drift counters stayed zero, every response was 200, and decode stats were normal. An instrument that asks *"did the parser succeed?"* cannot see *"did it succeed at producing garbage."*

## The fix

1. **`tool_required_keys_per_name()`** — a separate extractor from the allowlist one. Deliberately separate: the allowlist answers *"may this key appear?"*, the required set answers *"must it have appeared before the call may close?"*. Folding them is what hid this.
2. **`ToolGrammarXml`** tracks the emitted-key set per call, masks `</function>` until every required key has appeared, and rejects a duplicate key open. Both are dead states, so the decoder cannot sample them.
3. Threaded through `ToolConstrainer::begin()` and all four `server.cu` call sites.

### The part worth reviewing closely

The emitted-key set is part of **`signature()`**, because token legality now depends on it in two states: `KEY` (a duplicate is illegal) and `SLASH` (`</function>` is illegal until required are satisfied). `ToolMaskCache::get()` keys entirely on `signature()`, so omitting it makes two states with *different* `</function>` legality share one cached mask — which breaks in both directions depending on cache order: either a valid call can never close (hard wedge), or the required gate is silently bypassed (the original bug, back again). This is pinned by a test asserting the signatures differ *and* that `</function>` legality actually differs between them.

The set is monotone within a call, so this costs at most one extra cached state per parameter, not a combinatorial blow-up.

## Scope

**Degrades, does not break.** No `required` list, or no schema at all, keeps the previous shape-only behaviour — so Metal's 1-arg `begin()` and any future caller are unaffected. Optional parameters stay optional; parameter **order** stays free, because the schema says nothing about it.

**XML dialect only.** `ToolGrammar` (the JSON dialect) is not schema-aware at all — `reset()` takes only tool names and constrains `arguments` as generic valid JSON, so it will sample a missing required key, an **undeclared** key, or a duplicate. Measured, not assumed: all four shapes currently accept there. Closing it means constraining object keys at exactly one nesting level (the top of `arguments`, leaving nested object keys free) inside a generic JSON FSM that also handles escapes and `\uXXXX` in key strings — more machinery than the JSON path currently earns, and getting it subtly wrong would constrain keys it must not.

Rather than leave that asymmetry implicit, `--constrain-tools` now prints its dialect and the scope of its guarantee at startup:

```
constrain-tools: grammar-locked <tool_call> bodies (open=.. close=.. dialect=xml)
constrain-tools: enforcing required args + rejecting duplicate/undeclared param keys
```

or, on the JSON dialect:

```
constrain-tools: WARNING json dialect enforces SHAPE ONLY -- required args,
duplicate and undeclared keys are NOT constrained (XML dialect only)
```

An operator should learn the scope of the guarantee from the boot line, not from a client-side validation error three turns into a session.

## Testing

`test_toolconstrain` at 149 assertions, the new ones red before the change: the live shapes (missing required, `{}`, duplicate key) all refuse; valid calls in either parameter order pass; optional-omitted passes; no-schema and no-required stay permissive; and the mask cache key provably distinguishes two states whose `</function>` legality differs. Full `make test-tools` green from a clean build.

## Notes

- **Depends on #28**, which introduces `ToolGrammarXml`. Merge that first; this then applies to `master` cleanly.
- Worst case after the fix is an empty-but-present `path` — schema-valid, the client accepts it, the agent proceeds. Strictly better than `{}`.
- **Not verified:** no live re-run of the failing session (needs the GPU host), and neither backend compiled — no CUDA and no Apple-silicon machine available. Header-only C++ exercised by the CPU suites.

### DIFF
--- a/docs/BUILDLOG.md
+++ b/docs/BUILDLOG.md
@@ -14443,9 +14443,24 @@ client accepts it, and the agent can proceed. That is strictly better than
 
 The **JSON dialect** (`ToolGrammar`) is not schema-aware at all -- its
 `reset()` takes only tool names and constrains `arguments` as generic valid
-JSON, so it would emit `{"content": ...}` without `path`, and undeclared keys
-too. Same hole, other dialect. Left alone because the trained 3.8 format is
-XML and this needed to land on the reported failure first.
+JSON, so it will sample `{"content": ...}` without `path`, an undeclared key,
+or a duplicate key. Same hole, other dialect. Measured, not assumed: all four
+shapes accept on that path.
+
+DECIDED, not deferred by accident. The trained 3.8 emission is XML, which is
+where every observed failure came from. Closing the JSON hole means
+constraining object keys at exactly ONE nesting level -- the top of
+`arguments`, with nested object keys still free -- inside a generic JSON FSM
+that also has to handle `\uXXXX` and escapes in key strings. That is more
+machinery than the JSON path currently earns, and getting it subtly wrong would
+constrain keys it must not.
+
+So the reduced guarantee is made VISIBLE instead: `--constrain-tools` now
+prints its dialect and its scope at boot -- enforcing required args on XML, or
+`WARNING json dialect enforces SHAPE ONLY` -- and `ToolGrammar` carries the gap
+in a header comment. The whole lesson of this log is that a hole nobody can see
+survives; an operator should learn the scope of the guarantee from the startup
+line, not from a client-side validation error three turns into a session.
 
 Also still open: a parameter VALUE that legitimately contains `</tool_call>`
 truncates the call at the inner tag (the write-up documenting this very bug was
--- a/docs/BUILDLOG.md
+++ b/docs/BUILDLOG.md
@@ -14350,3 +14350,116 @@ is a source-level fact about the vendored header's API, and the CUDA-side
 changes are header-only C++ exercised by the CPU suites, but
 `build/q27-server` and `q27-metal-server` both still want a real compile before
 anyone leans on this.
+
+## 2026-08-22 (b): `--constrain-tools` guaranteed the SHAPE and nothing else -- schema-invalid calls, 200 OK, zero drift logged
+
+Filed as issue #2 against `da3d2fc`, from a live agentic session on
+`qwen38-27b-mtp` with `--constrain-tools`. The agent was asked to write a file.
+Every attempt came back from the CLIENT as schema-invalid:
+
+```
+Validation failed for tool "write":
+  - path: must have required properties path, content
+  Received arguments: {}
+```
+
+and the server log for the same turn was clean:
+
+```
+[toolgram] engaged (rem=0, dialect=xml)
+<function=write>
+<parameter=content>
+...
+</parameter>
+</function>
+</tool_call>
+[toolgram] call closed
+```
+
+`<parameter=path>` was never emitted, and the grammar **accepted the close
+anyway**. The file stayed 0 bytes across every retry.
+
+### What the guarantee actually covered
+
+`ToolGrammarXml` shipped as "schema-aware", and it was -- for the ALLOWLIST.
+`tool_param_keys_per_name()` extracts `properties`, so the grammar knows which
+keys *may* appear. `parameters.required` was parsed **nowhere**: the only
+consumer of `required` in the tree was `infer_tool_name()`, the post-hoc
+mode-6 rescue that guesses which tool an orphaned args object belongs to --
+after-the-fact selection, not decode-time enforcement. So the grammar had no
+notion of which keys *had* appeared, `DONE_` accepted `</function>` on any
+subset including the empty one, and `KEY`->`GT2` let the same key open twice.
+
+Two failure classes fell straight through a constraint advertised as a hard
+correctness guarantee: **missing required arguments** and **duplicate keys**.
+The duplicate one compounded in the live session because the file being written
+was itself a tool-call verifier -- its content quoted `</parameter>
+<parameter=content>`, the model emitted those as real tags mid-value, and the
+grammar accepted the second `content` key, doubling the value.
+
+### Why every safety net stayed quiet
+
+This is the fourth time in this log, and the sharpest instance. `[drift]
+UN-RESCUED` and `Q27_DRIFT_CORPUS` fire when the FALLBACK PARSER fails. These
+calls parsed perfectly -- they were well-formed XML naming a declared tool. So
+the corpus stayed empty, the drift counters stayed zero, every response was
+200, and the speculative/decode stats were normal. **The only component in the
+system that knew anything was wrong was the client's JSON-schema validator, on
+the other side of the wire.** An instrument that reports "did the parser
+succeed?" cannot see "did the parser succeed at producing garbage."
+
+Worth stating plainly: constraining the SHAPE of a call and constraining its
+VALIDITY are different guarantees, and the config claiming the former as the
+latter is how this shipped.
+
+### The fix
+
+`required` is threaded to where the decision is made. `ToolGrammarXml` now
+tracks the emitted-key set per call, masks `</function>` until every required
+key has appeared, and rejects a duplicate key open. `tool_required_keys_per_name()`
+is a separate extractor from `tool_param_keys_per_name()` deliberately -- the
+allowlist answers "may this key appear?", the required set answers "must it
+have appeared before the call may close?", and folding them hid that they are
+different questions.
+
+The emitted set is part of `signature()`, because token legality now depends on
+it in two states: `KEY` (a duplicate is illegal) and `SLASH` (`</function>` is
+illegal until required are satisfied). Omitting it would let a cached mask
+authorise a close the grammar must refuse. The set is monotone within a call,
+so this costs at most one extra cached state per parameter rather than a
+combinatorial blow-up.
+
+DEGRADES, does not break: no `required` list (or no schema at all) keeps the
+previous shape-only behaviour, so callers that lack the schema -- Metal's 1-arg
+`begin()`, any future caller -- are unaffected. Optional parameters stay
+optional, and parameter ORDER is still free, because the schema says nothing
+about it.
+
+Worst case after the fix is an empty-but-present `path`: schema-valid, the
+client accepts it, and the agent can proceed. That is strictly better than
+`{}`.
+
+### Not fixed here, recorded
+
+The **JSON dialect** (`ToolGrammar`) is not schema-aware at all -- its
+`reset()` takes only tool names and constrains `arguments` as generic valid
+JSON, so it would emit `{"content": ...}` without `path`, and undeclared keys
+too. Same hole, other dialect. Left alone because the trained 3.8 format is
+XML and this needed to land on the reported failure first.
+
+Also still open: a parameter VALUE that legitimately contains `</tool_call>`
+truncates the call at the inner tag (the write-up documenting this very bug was
+itself cut at 651 bytes by it). Needs the TOOL channel to close only on a
+balanced dialect.
+
+### Measured
+
+`test_toolconstrain` 149 assertions, the new ones red before the change: the
+live shapes (missing required, `{}`, duplicate key) all refuse; valid calls in
+either parameter order still pass; optional-omitted passes; no-schema and
+no-required stay permissive; and the cache key provably distinguishes two
+states whose `</function>` legality differs. Full `make test-tools` green from
+a clean build.
+
+NOT VERIFIED: no live re-run of the failing session (needs the GPU host), and
+neither backend compiled -- no CUDA and no Apple-silicon machine available.
--- a/docs/BUILDLOG.md
+++ b/docs/BUILDLOG.md
@@ -14187,3 +14187,166 @@ test instead of being folklore.
 
 Green on all three targets: CPU suites under gcc on haight and clang on the M4,
 q27-server complete under nvcc, q27-metal-server clean under -Werror.
+
+## 2026-08-21: the hallucinated-result guard was vetoing REAL calls -- presence is not enclosure
+
+Follow-on to (d). The report was the same sentence as always -- "sometimes I
+still see `<parameter=` / `</function>` in the output" -- but this time the
+bytes recovered cleanly when replayed in isolation and died in the session. The
+difference was the CONTEXT they arrived in, not the call.
+
+**The shape.** A perfectly recoverable mode-21 openerless call, in a tail that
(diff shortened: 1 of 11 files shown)


kaushikvira:
## ✅ Build + tests passed

Tested the exact PR head `298397b3ad8fc798d87ecb93cc99613f92b8b2d4` on a local checkout (g++ -O2 -std=c++17, RTX 5090 host).

| Suite | Result |
|-------|--------|
| test_openai_bridge | PASS |
| test_think_resolve | PASS |
| test_stream_split | PASS |

signalnine:
could you rebase this on current master when you get a chance? and the conflict
is my doing, not yours -- you specified #28 first and i merged #31 and #32 ahead
of it, so this now collides in `docs/BUILDLOG.md` and `tools/test_tool_drift.cpp`.
both are append-order collisions rather than logic.

master is now at 6412204 and carries #26, #29, #30, #31, #32 plus a follow-up of
mine (below).

on the #28 commit this carries: i verified it before #31/#32 went in and it holds
up, so the rebase is the only thing blocking it. what i checked, so you don't
have to re-do it:

```
unclosed <result> enclosing the call          REFUSE   (master and #28)
<output> inside the span                      REFUSE   (master and #28)
closed then reopened <result>                 REFUSE   (master and #28)
mode22 shape enclosed by unclosed <result>    REFUSE   (master and #28)
closed <output> BEFORE a real call            REFUSE on master -> ACCEPT on #28
```

only the false veto flips, which is the whole point. the depth counting reads
correctly to me too: `i` strictly advances and a stray close won't go negative.

your span-underflow fix is a real crash, not just a tidy-up. i reproduced it on
master before your PR existed:

```
call[0] Read  source_begin=57    source_end=NPOS
call[1] Read  source_begin=NPOS  source_end=229
*** THREW: basic_string::substr: __pos (18446744073709551615) > size (229)
```

that's `std::out_of_range` escaping into the request handler on any mode-20
batch. my own tests missed it because they exercised `parse_bare_tool_calls`
directly and never drove a multi-call batch through `append_text`. so thanks for
that one.

also worth telling you: #29 was necessary but not sufficient. 0.53.1 ALSO removed
`Request::is_connection_alive` in favour of `is_connection_closed`, and inverted
the sense -- it defaults to `[]{ return true; }` meaning closed, while the metal
handlers take a liveness predicate. a straight rename compiles clean and reports
every live connection as dead. fixed in 3635573 by negating at both sites, and
q27-metal-server now actually builds under -Werror on apple clang 21. i found it
by compiling rather than grepping, which is the difference.

kaushikvira:
## Rebased onto current master + tests green

Rebased onto `08c1680` (master was `6412204` when you asked; one more commit of yours, `parser: recover every call in a wrapped batch`, landed since). The PR is now its own 3 commits on top of master.

**Conflicts resolved** (3 files, 1 real merge + 2 append-order):

- `src/api_common.h` — your batch rewrite of the mode-16 `<function...>` opener path collided with #28's switch of that mode to `hallucinated_result_around`. Resolved as **both**: kept your batched loop and added the per-span hallucination guard to it (a `<result>`/`<output>` enclosing or inside a candidate span breaks the batch rather than promoting fiction to a call), consistent with #28's intent that mode 16 sit on the shared helper.
- `docs/BUILDLOG.md`, `tools/test_tool_drift.cpp` — pure append-order; both sides kept, both new test functions registered (`test_repeated_opener_is_not_shown_as_text` and `test_hallucinated_result_enclosure_rule`).

**Tested on rebased head `9fd3dc4`** (g++ 15.2 -O2 -std=c++17, CPU-only host — no CUDA/Apple-Silicon here):

`make test-tools`: **exit 0**

| Suite | Result |
| --- | --- |
| test_openai_bridge | PASS |
| test_think_resolve | PASS |
| test_stream_split | PASS |

CPU `make all` targets also compile clean. As before, neither backend binary was compiled on this host (no CUDA / no Apple-silicon), so `q27-server` and `q27-metal-server` still want their respective real compiles.

Still draft — say the word and I'll flip it to ready.

kaushikvira:
Using this build from last 20 minutes, not a single issue with tool calls. so seem working good.

kaushikvira:
File Edit sometime still acting up, means something is still missing. Lot better but still not 100%  :(

signalnine:
merged, thanks -- and thanks for closing #28 yourself, that was the right call.
resolving the api_common.h collision as BOTH rather than picking a side was
exactly right, and i checked it holds: the live 14-call failure still recovers
all 14 with your per-span guard applied, and all four invented-result shapes
still refuse with only the false veto flipping.

built the two binaries you couldn't: q27-server complete under nvcc with
ToolGrammarXml linked, q27-metal-server clean under -Werror on apple clang 21,
CPU suites green under gcc and clang. 174 assertions.

> File Edit sometime still acting up

found one cause, and Edit is the only tool that would ever notice. its oldString
has to match the file byte for byte, so a value the parser rewrites fails to
apply while the same rewrite is invisible to Bash and Read. that is also why it
was intermittent -- it breaks exactly when the edited region happens to end in
whitespace.

the value extractor consumed one leading newline (right, the dialect puts the
value on its own line) and then greedily trimmed every trailing '\n' and ' '.
round-trips through the dialect before the fix:

```
"line1\n"      ->  "line1"      a block ending in a blank line
"code    "     ->  "code"       a line with trailing whitespace
"   "          ->  ""           a whitespace-only value
```

each of those is a failed Edit. fixed in e1bbde7: the trailing side now mirrors
the leading side, exactly one delimiter newline plus its \r for CRLF. full suite
green, so nothing was relying on the greedy trim.

one more thing i found and deliberately did NOT fix in the same pass: a value
that CONTAINS `</parameter>` truncates there.

```
"before\n</parameter>\nafter"  ->  "before"
```

same collision you solved for `</tool_call>` in #31, and the same shape of fix
applies: the real closer is the last `</parameter>` before the next
`<parameter=` or `</function>`. i left it because it is a second change to the
same function with no evidence it is what you hit -- your edits are Go files,
which don't contain the dialect. worth its own PR if you want it, and your
provisional-closer reasoning is the model for it.

if Edit still misbehaves after e1bbde7, the most useful thing would be the raw
oldString the model emitted versus what the file actually contains -- the
difference will point straight at whichever byte we are still eating.

PR #36, opened 2026-08-22
toolgram XML: re-engage on bare <function=, fix on_pending_xml guards (#35)

### LINKED ISSUE #35: toolgram XML: one corrupt byte in a param key disengages the constraint and loses the whole tool call
## Symptom

Tool calls corrupt mid-flight and come back to the client as plain text in a 200:
the model emits a spurious duplicate parameter whose KEY carries junk bytes
(`<parameter=edits!!`), the engine catches it too late, then the whole call is
lost. Observed 4 disengages / 3 UN-RESCUED calls in one live session
(host log `/tmp/q27-server.log`, binary `build/q27-server --constrain-tools`,
Qwen38-27B, XML dialect).

## Evidence chain (live log)

```
[toolgram] engaged (rem=0, dialect=xml)
[toolgram] disengaged: byte 0x21 rejected        <- 0x21 = '!' (see repro)
[drift] UN-RESCUED (ntools=14) intended tool call: <function=edit>
<parameter=edits>
[{"newText": ...
```

## Root cause

Three stacked problems:

1. **The '!' should never have been sampleable.** Under an active, schema-aware
   `ToolGrammarXml`, KEY state must reject `!` (not a prefix of any declared
   param). That it reached the stream means the `!`-bearing token was sampled
   WITHOUT the current grammar mask — a gap between mask construction
   (toolgram.h) and mask enforcement at decode time, so the veto only fires at
   feed-time (`on_id`), i.e. after the byte is already committed.

2. **On a rejected byte, `drop()` disengages the constraint for the rest of the
   call** (`toolconstrain.h`), so the model free-generates the remainder
   (`": {}</parameter>...`) unconstrained. One bad byte permanently kills the
   guarantee; no recovery/re-engage is attempted.

3. **The drift parser cannot rescue the result.** `edits!!"` as a parameter key
   matches no drift mode (14-22), so `[drift] UN-RESCUED` and the client sees
   text where a tool call should be.

## Repro (deterministic, hermetic)

`tools/repro_edits_bang.cpp` — feeds the exact byte sequence through
`ToolGrammarXml` (schema `{edits,path}`, required `path`):

```
DEAD at byte 133 = 0x21 ('!')
  context: ...ameter>
<parameter=edits!
  prefix up to byte 132 accepted: yes
```

Pure CPU, byte-deterministic: grammar accepts the valid prefix, dies at the
first '!'. (The model actually *emitting* the junk is probabilistic, but it
occurred 4x in one session.)

## Suggested fix direction

- (a) Root: enforce the mask at sample time so a grammar-illegal token is never
  sampled in the first place (audit the constrained-sampling path / draft
  accept).
- (b) Recovery: on mid-call dead state, re-engage the XML grammar at the next
  `<function=` in the same body rather than disengaging for the whole call.
- (c) Parser: teach `parse_bare_tool_calls` the spurious-parameter shape
  (`<parameter=KEY<garbage>>`) so a reachable call still survives server-side.

Regression test from (repro) will be added to the test suite in the PR.

kaushikvira:
Fixes signalnine/q27#35.

## What broke

The 3.8 model opened a spurious second `<parameter=edits!!` in a `<function=edit>`
call. The schema-aware `ToolGrammarXml` rejects the '!' at KEY state, but:
1. the '!' reached the stream at all (sample-time gap, tracked in #35),
2. `drop()` then disengaged the constraint for the rest of the call, and
3. `parse_bare_tool_calls` had no drift mode for the `edits!!"` key shape, so
   `[drift] UN-RESCUED` and the call came back to the client as text in a 200
   -- the agent lost the tool call.

Live evidence: 4x `[toolgram] disengaged: byte 0x21 rejected`, 3x
`[drift] UN-RESCUED` in one session.

## This PR (host-side, CPU-gated)

1. **`scan_round` re-engages the XML grammar on a bare `<function=` opener --
   DEFAULT ON** (only for the XML dialect): the wrapper-less 3.8 drift class AND
   the remainder of a body whose mid-call reject dropped the constraint. The
   wrapper-only default was exactly what let the observed corruption through.
   `Q27_TG_REENGAGE=0` restores the old wrapper-only behaviour (A/B escape
   hatch). Mirrors the `<tool_call>` trigger exactly (complete-within-token,
   reset+feed rem, stage, truncate). **XML-specific by design** -- JSON
   wrapper-less drift is a bare `{...}` whose '\{' trigger would constrain
   JSON-in-prose, so JSON stays wrapper-only and is recovered at parse time.
   False-positive blast radius is bounded and self-limiting: auto-disengage at
   the first non-conforming byte, emitted bytes are never rewound, so prose
   quoting the dialect is only steered if it forms a full well-formed call.
2. **`on_pending_xml` missing guards fixed** -- an XML-PATH-ONLY bug: the JSON
   path (`on_pending`) already had `enabled/active/id` guards; the XML path
   could stage + ACTIVATE a mask while inactive by advancing the stale
   reset-state grammar over a pending token that merely looks like
   `<function...>`.
3. **Regression tests** in `tools/test_toolconstrain.cpp`: the exact live byte
   sequence dies at byte 133 ('!'), bare-`<function=` re-engages and constrains
   to completion (default-on, and no-op under `=0`), and inactive
   `on_pending_xml` no longer arms a mask.

## Verification

- `make test-tools`: 0 FAIL (incl. 3 new tests)
- `q27-server` rebuilds clean under nvcc (CUDA 13.3)
- Kernels untouched (host-side only); GPU gates not run here because the
  serving 5090 is fully occupied -- can be run off-hours on a free card.

The one change from the original plan: re-engage went from opt-in to
default-on after review -- wrapper-only engagement was the default that let the
bug through, and the env-gated A/B showed the only risk (constraining prose
that quotes the dialect) is self-limiting.

### DIFF
--- a/src/toolconstrain.h
+++ b/src/toolconstrain.h
@@ -19,6 +19,7 @@
 
 #include <cstdio>
 #include <cstdlib>
+#include <cstring>
 #include <string>
 #include <vector>
 
@@ -33,6 +34,29 @@ struct BasicToolConstrainer {
     std::vector<int>* host2dev = nullptr;
     bool enabled = false, active = false;
     bool pool_dead = false; // sticky: mask pool filled up this request
+    // Q27_TG_REENGAGE (XML dialect): re-engage the grammar on a bare
+    // <function= opener (no <tool_call> wrapper) during scan_round -- the
+    // wrapper-less 3.8 drift class AND the remainder of a body whose mid-call
+    // reject just dropped the constraint (signalnine/q27#35). DEFAULT ON for
+    // the XML dialect: wrapper-only engagement was the default that let the
+    // observed corruption through. False-positive blast radius is bounded and
+    // self-limiting -- engagement auto-disengages at the first non-conforming
+    // byte and emitted bytes are never rewound, so prose that merely QUOTES
+    // the dialect (<function=...> in an answer) is only ever steered if it
+    // forms a full well-formed call, and simply disengages otherwise. Set
+    // Q27_TG_REENGAGE=0 to restore the old wrapper-only behaviour (the A/B
+    // escape hatch); "1"/unset selects the default. XML-specific by design:
+    // the XML wrapper-less drift form is a distinctive tag, while the JSON
+    // drift form is a bare {...} whose "{" trigger would constrain JSON the
+    // model writes in prose (configs, code samples) -- so JSON wrapper-less
+    // drift is recovered at PARSE time (api_common.h drift modes) instead,
+    // and the JSON path stays wrapper-only here.
+    // NOTE (scope, signalnine/q27#35): this is RECOVERY + correctness only --
+    // it cannot stop a byte that the grammar would reject from being
+    // SELECTED and emitted under a stale/host mask (the '!' sample-time gap).
+    // That gap is tracked separately in #35; re-engage just makes a dropped
+    // bare/wrapper-less call constrain properly once its opener is seen.
+    bool reengage_bare = false;
     ToolGrammar tg;
     ToolGrammar staged_state; // grammar state whose mask is in verify slot 0
     ToolGrammarXml staged_state_xml; // parallel for the XML dialect (P11 on_drafts)
@@ -52,6 +76,7 @@ struct BasicToolConstrainer {
         tail.clear();
         names = std::move(n);
         dialect_xml = false;
+        reengage_bare = false;
         params_per_name.clear();
         required_per_name.clear();
     }
@@ -78,6 +103,9 @@ struct BasicToolConstrainer {
         dialect_xml = dialect_xml_;
         params_per_name = std::move(pp);
         required_per_name = std::move(rq);
+        // default ON for XML; Q27_TG_REENGAGE=0 opts out (A/B hatch)
+        const char* e = getenv("Q27_TG_REENGAGE");
+        reengage_bare = dialect_xml_ && (!e || strcmp(e, "0") != 0);
         if (dialect_xml_) tg_xml.reset(names, params_per_name, required_per_name);
     }
     // pool id for grammar state g's legal-token mask (-1 if pool full)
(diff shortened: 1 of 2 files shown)


signalnine:
merged, thanks. the repro is what made this quick to take -- feeding the exact
byte sequence through ToolGrammarXml beats any amount of description.

checked the blast radius before merging rather than taking it on faith:
--constrain-tools defaults to false and scan_round returns early when disabled,
so none of this runs on the default serving path. within the opt-in, i like that
re-engage self-limits -- auto-disengage at the first non-conforming byte and no
rewind of emitted bytes means a false positive costs a few tokens of steering
and nothing else. the Q27_TG_REENGAGE=0 hatch is the right call for something
that changes when the grammar engages.

the on_pending_xml guards are the part i'd have missed. staging AND activating a
mask off a stale reset-state grammar, because the JSON path grew the guards and
the XML path didn't, is exactly the kind of asymmetry that only shows up when
something else goes wrong first.

**#35 stays open**, and i've said so in the merge commit. this makes the failure
recoverable, it doesn't make it impossible -- the root cause is still that the
'!' was sampleable, because the mask is enforced at feed time, after the byte is
committed. worth its own change, and probably a bigger one: it's the difference
between a grammar that vetoes and a grammar that constrains.

one thing i could NOT verify: metal. `src/metal/metal_server.cpp` includes
`../toolconstrain.h` and compiles it under -Werror, and my mac is unreachable
right now. the diff has nothing clang-sensitive in it, but that's an argument
rather than a build, so it's recorded as unverified in the merge commit and i'll
confirm when the machine is back. that exact gap bit us on the httplib bump, so
i'd rather flag it than let it ride quietly.

full CPU suite green here, q27-server complete under nvcc, and the merge keeps
this week's parser work (your branch point is old, so the raw tree diff looks
alarming and isn't).

signalnine:
metal gate closed: `q27-metal-server` links clean under `-Werror` on apple clang
(macOS 26.5, arm64) and the CPU suites pass there too. so this is now verified on
all three targets.

worth noting it was the right thing to hold open: PR #37, reviewed in the same
sitting, DOES break that target -- it adds a member to `Msg` and
`metal_server.cpp` brace-initialises it at eight sites, all hard errors under
`-Wmissing-field-initializers`. `make test-tools` and the CUDA build are green on
that tree. same blind spot you called out on #29.

Issue #2, opened 2026-07-17
Title: License intent + interest in a Metal backend merge-back?


manthedan:
Hey @signalnine! I found your project looking to see if anyone had made a DS4 for Qwen, great job! 

I've created a fork https://github.com/manthedan/q27 (branch metal) that adds an Apple Silicon / Metal backend, and support for the [Bonsai-27B](https://prismml.com/news/bonsai-27b) quants for those of us with lower memory Macs.

1. What do you intend the license to be? There's no LICENSE file in the repo. Is the intent MIT, like ds4? I'd like to distribute prebuilt binaries of the Metal fork (Homebrew tap), and I don't want to do that without knowing your terms — everything I've built inherits from your code. Happy to send a one-line PR adding the LICENSE file once you say which one.

2. Are you interested in the Metal work merging back, so q27 stays canonical and supports both CUDA and Metal the way ds4 covers its targets? Most of my work is self-contained under src/metal/ (engine, shaders, server, CLI) plus format additions (ternary/binary weight dtypes) and docs. I wouldn't propose that as one PR. If you're interested, I'd suggest incremental: agree on a backend split first, then land it in reviewable pieces. If you'd rather keep q27 CUDA-only and single-target, no hard feelings — I'll keep the fork clearly labeled as a downstream Metal port and keep pulling from you.

Appreciate the hard work!

signalnine:
Awesome! Good catch, I completely forgot to add a LICENSE file. It's going to be MIT, I'll add that shortly here. And yes, I'd love your Metal work to merge back into this project.

Issue #1, opened 2026-07-15
Guidance wanted: maximizing context on ~22GB usable VRAM (A10G, less than the 24GB turbo3 numbers in BUILDLOG)


chaudhryfaisal:
## Summary

Running Qwen3.6-27B-MTP on a cloud A10G instance with `Q27_KV=turbo3` already enabled, and reaching
~28K context. `docs/BUILDLOG.md` documents turbo3 reaching ~131K context on a 24GB card — my
instance only has ~22GB usable, not the full 24GB, so I'm not assuming anything is broken, just
looking for guidance on the best way to approach the smaller budget and get as much context as
this card can realistically support.

## Environment

- GPU: NVIDIA A10G, 24GB nominal, compute capability sm_86 (same generation as 3090/3080)
- VRAM reported by `nvidia-smi`: ~22 GiB usable
- Cloud provider: cloud GPU instance (not a local card)
- q27 build: default build
- Model file: Qwen3.6-27B-MTP, v1.4 policy (~17.7GB on disk)
- KV cache setting: `Q27_KV=turbo3`

## Current behavior

- Server boots and serves normally with `Q27_KV=turbo3` set.
- Reaching ~28K context on this instance, vs. the ~131K figure documented for a full 24GB card.

## What I'd appreciate guidance on

Given ~2GB less headroom than the 24GB case in BUILDLOG, I'd naively expect a modest reduction in
max context, not a drop from ~131K to ~28K — turbo3 KV looks cheap enough per token that 2GB less
shouldn't cost that much on its own, so I may be missing something about how the fixed costs
(weights, CUDA graphs, scratch, slot count) scale down on a tighter budget. Rather than guess, I
wanted to ask directly:

- Is `build/q27-server-w8` (`Q27_W_MAX=8`) the right build target to try for a card in the
  ~22GB-usable range, or is there a tighter setting worth trying below W8?
- Are there other flags (slot count, scratch sizing, graph capture options) worth adjusting
  specifically when working with less than the full 24GB, beyond just setting `Q27_KV=turbo3`?
- Is there a way to see what the fixed cost (weights + graphs + scratch) is actually measuring at
  on this card, so I can tell how much of the ~22GB is fixed overhead vs. available for KV cache?

Happy to run whatever combination of flags/build settings would help and report back exact numbers.

## Thanks

Really appreciate the amount of engineering and the sheer amount of measured, written-down
groundwork

signalnine:
[...]

PR #37, opened 2026-08-22
Port froggeric Qwen3.8 chat template (v22.3) behaviors into the internal renderer

PierpaoloPernici:
## Intent

This PR aligns q27's chat-template renderer with **froggeric's Qwen 3.8 chat template (v22.3)** — [froggeric/Qwen-Fixed-Chat-Templates](https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates) — with the stated goal of achieving **the best integration with agent harnesses** (Claude Code and pi agent in particular): correct history rendering (including reasoning), effective thinking/effort control, robust tool handling, and no context loss across multi-turn loops.

The "v22"/"v22.3" references in the commits refer to the **froggeric template** version, not to a q27 version.

### Upstream reference
Every behavior ported here is documented on the template's model card ([froggeric/Qwen-Fixed-Chat-Templates](https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates)) — this PR is a C++ port of those documented behaviors, not a new design:

| This PR | Documented on the model card |
|---|---|
| ~~Item 1 — default effort `medium~~ **NOT ported (dropped per review)** — q27 keeps its own default; see [Thinking / effort control](#thinking--effort-control) |
| P0 reasoning preservation / dedup | Table row **History Reasoning Extraction** + changelog v22.3 #2 (*Reasoning De-duplication*) |
| E — unconditional think wrap | Kwarg reference §2 (**`preserve_reasoning` & `preserve_thinking`**, default `true`) |
| B — multi-system merge | Table row **Leading System Prompts** (*joined by double newlines*) + changelog v22.2 #3 |
| P1a — inline `<\|think_*\|>` toggles | Technical details §2 (**Upfront Pre-Scan for Control Tags**) + changelog v22.1 #3 |
| Case-insensitive aliases (`high`/`max`/`ultracode`/`extreme` → `xhigh`, …) | Kwarg reference §1 (**API Compatibility Aliases**) + changelog v22.1 #4, v22.2 #2 |
| P1b — two-tier error escalation | Table row **Tool Error Recovery** + technical details §5–**§6 Smart False-Positive Detection** (120-char head, strong vs weak signals) + changelog v22.3 #1 |
| Item 6 — XML tool-call dialect | Technical details §4 (**Native XML Tool Format**) |
| P3 / items 7/8 — truncation, JSON bypass | Kwarg reference §3–§4 (**Dynamic Payload Truncation**: *"Both limiters are disabled automatically when `tool_call_format="json"`"*) |
| Item 2 — `none`/`off` disables thinking | Table row **Non-Reasoning Fast Mode** (*`reasoning_effort="none"`*) + Kwarg §1 |

Template credits: fixes by [Frédéric Guigand (@froggeric)](https://huggingface.co/froggeric); error tiering and test suite by [Juan Calderon-Perez (@g-a-b-y)](https://huggingface.co/g-a-b-y). Apache-2.0.

---

## What it does

The internal renderer (`chatml_prompt`/`api_common.h`) is native C++ (not minijinja): this PR brings the template's **key behaviors** directly into the code, with no added dependencies.

### History rendering
- **P0 — Reasoning preservation (OpenAI path)**: `reasoning_content` on assistant messages is now read on input and rendered as a `<think>...</think>` block in the prompt instead of being silently dropped. Scope correction per review: the Anthropic path was NOT broken — `api_common.h` already re-rendered `thinking` blocks into history (`content = "<think>\n" + think + ...`). The real gap was the OpenAI shape, where `reasoning_content` as a sibling field was never read; this patch adds exactly that.
- **P0 — Dedup**: if the assistant content opens with a think block (`<think>`/`<thinking>`, or an orphan closer `</think>`/`</thinking>`), the leading block is removed before rendering (no double render). With item E the strip is unconditional on every assistant turn: even without explicit `reasoning`, the inline block no longer duplicates the rendered one.
- **E — Unconditional wrap**: every assistant history turn is wrapped in `<think>...</think>` (even empty), matching the template's `preserve_thinking=true`.
- **B — Multi-system merge**: consecutive leading `system`/`developer` messages are merged into a single block (joined with a blank line).

### Thinking / effort control
- **P1a — Inline toggles** `<|think_off|>`, `<|think_on|>`, `<|think_low|>`, `<|think_minimal|>`, `<|think_medium|>`, `<|think_xhigh|>`, `<|think_high|>`, `<|think_ultracode|>`, `<|think_extreme|>`, `<|think_max|>`: scanned before rendering with stateful tracking across messages, stripped from the prompt, and they drive both the `reasoning_instructions` line and the generation prompt.
- **Item 1 — default effort `medium`: DROPPED per review.** The effective default stays `xhigh`/`off` (per boot dialect), so every existing client and benchmark baseline is unchanged. `medium` remains reachable explicitly via `Q27_REASONING_EFFORT=medium`, the `reasoning_effort` request field, or a `<|think_medium|>` tag. Upstream parity for this item is deferred until a measured benchmark arm justifies it (the n=3 arm from 2026-08-19 — mean 0.545 xhigh vs 0.453 medium — supports neither side).
- **Item 2 — `reasoning_effort: none/off` disables thinking** end-to-end (prompt + engine).
- **Case-insensitive aliases**: `high`/`max`/`ultracode`/`extreme` → `xhigh`; `minimal` → `low`; `none`/`off` → disabled.
- **Honored request fields** (with `--request-think`): `reasoning_effort`, `output_config`/`reasoning` (`.effort`/`.output_effort`, `.disabled`/`.enabled`, `.budget_tokens`, `.output_budget_tokens`), top-level `budget_tokens`, `thinking.type`, `enable_thinking`.

### Tools
- **P1b — Two-tier error escalation**: tracks consecutive tool-response failures and injects the system warnings (1st "diagnose and retry", 2+ "fundamentally different approach").
- **P1b — Switch, default OFF** (per review, second round): the escalation is now opt-in — `Q27_TOOL_ERROR_WARNINGS=1` enables it at boot; a per-request boolean `tool_error_warnings` (top-level or `chat_template_kwargs`) overrides either way. When off, failure detection is skipped entirely and no warning is ever injected. Template parity stays one flag away; flipping the default back would require a measured arm showing the escalation helps.
- **Known FP profile** (measured with a direct probe of `is_tool_response_failure`, since session logs are not persisted): over realistic benign outputs, 3/12 false-positive — grep/diff exit code 1 (no-match/differences convention) and short `error: none` summaries fire; recall on true failures is 6/6. See the review reply for details.
- **P1b — Smart false-positive detection** (v22.3): 120-char head, code/grep output suppression (`throw new`, `console.error`, etc.), `exit code` handling (0 = not an error), strong vs weak error.
- **P2 — Grouping**: consecutive tool responses grouped into a single `user` turn.
- **Item 6 — Tool calls in the active dialect**: assistant tool calls in history rendered in XML per-parameter (Qwen's native dialect) or JSON.
- **P3 / items 7/8 — Truncation**: `max_tool_arg_chars` / `max_tool_response_chars` (env or request fields), bypassed in the JSON dialect.

---

## What is NOT included
- **Vision** (`image_pad`/`video_pad`): the server has no vision pipeline.
- **`tool_call_format` per-request**: the constraint grammar and drift parser are boot-level (continuous batching = a single dialect), so the format is selected at boot (`Q27_TOOL_DIALECT` or auto-detect).
- **`preserve_reasoning`/`preserve_thinking` as a request field**: default `true` (current behavior); the off switch is not exposed.

---

## Tests
- New regression tests in `tools/test_openai_bridge.cpp` and updates to `test_tool_drift`/`test_tokenizer`.
- All suites pass (`test_openai_bridge`, `test_chat_completions_integration`, `test_think_resolve`, `test_tool_drift`, `test_toolconstrain`, `test_tokenizer`).
- Verified at runtime with the server (`--think --request-think`): `reasoning_effort none/off` → `reasoning_tokens 0`; `budget_tokens` → cap applied; `output_config.effort` → 200; `reasoning_content`/`thinking` preserved and recalled by the model.

---

## Notes for the reviewer
- **RESOLVED — default effort**: the `xhigh`→`medium` default change originally in `449bee2` has been reverted in `a10246b` per review; the effective default is unchanged from master. No observable behavior change remains around the default.
- **Metal build**: fixed in `7058263` — `Msg::reasoning` now carries an NSDMI (`std::string reasoning{};`) so the eight 2-field brace-init sites in `metal_server.cpp` stay valid under `-Werror -Wmissing-field-initializers`.
- The server must be started with **`--think --request-think`** for the request fields (effort, budget, disabled) to be honored.
- The P0–P3 commits and the v22.3 items are listed individually: the series starts with the port and continues with the v22.3 refinements.

### DIFF
--- a/src/api_common.h
+++ b/src/api_common.h
@@ -313,6 +313,39 @@ inline std::string reasoning_effort_line() {
     return reasoning_effort_line_level(reasoning_effort_env_level());
 }
 
+// v22.3 request-driven template options (items 3/4/5).
+// effort: -1 = env/boot default, else 0 medium, 1 low, 2 xhigh.
+// auto_disable_thinking_with_tools: start thinking OFF when tools are present.
+// tool_format: -1 = boot dialect, 0 json, 1 xml.
+struct TemplateOpts {
+    int effort = -1;
+    bool auto_disable_thinking_with_tools = false;
+    int tool_format = -1;
+};
+inline int effort_string_level(std::string v) {
+    for (auto& c : v) c = (char)tolower((unsigned char)c);
+    if (v == "minimal" || v == "low") return 1;
+    if (v == "xhigh" || v == "high" || v == "max" || v == "ultracode" || v == "extreme") return 2;
+    return 0; // none/off/medium/unknown
+}
+inline TemplateOpts template_opts_from_body(const json& body) {
+    TemplateOpts o;
+    auto read = [&](const json& b) {
+        if (b.contains("reasoning_effort") && b["reasoning_effort"].is_string())
+            o.effort = effort_string_level(b["reasoning_effort"].get<std::string>());
+        if (b.contains("auto_disable_thinking_with_tools") && b["auto_disable_thinking_with_tools"].is_boolean())
+            o.auto_disable_thinking_with_tools = b["auto_disable_thinking_with_tools"].get<bool>();
+        if (b.contains("tool_call_format") && b["tool_call_format"].is_string()) {
+            const std::string f = b["tool_call_format"].get<std::string>();
+            o.tool_format = (f == "json") ? 0 : (f == "xml" ? 1 : -1);
+        }
+    };
+    read(body);
+    if (body.contains("chat_template_kwargs") && body["chat_template_kwargs"].is_object())
+        read(body["chat_template_kwargs"]);
+    return o;
+}
+
 // P1a (2026-08-22): the <|think_*|> toggle markers from the qwen3.8
 // chat_template. Stripping: these are BPE-tokenisable control markers the
 // model is trained to read in user/system content; they must not leak into
@@ -335,8 +368,9 @@ struct ThinkToggles {
     bool thinking = true;
     int effort = 0; // 0 medium, 1 low, 2 xhigh
 };
-inline ThinkToggles scan_think_toggles(const std::vector<Msg>& msgs, bool base_think) {
-    ThinkToggles st{base_think, reasoning_effort_env_level()};
+inline ThinkToggles scan_think_toggles(const std::vector<Msg>& msgs, bool base_think,
+                                       int base_effort = -1) {
+    ThinkToggles st{base_think, base_effort >= 0 ? base_effort : reasoning_effort_env_level()};
     for (const Msg& m : msgs) {
         if (m.role != "system" && m.role != "user") continue;
         if (m.role == "user" && m.content.rfind("<tool_response>", 0) == 0) continue;
@@ -486,7 +520,8 @@ inline std::string chatml_prompt(const std::vector<Msg>& msgs, const json& tools
                                  bool think = true, size_t* stable_off = nullptr,
                                  size_t* sys_off = nullptr,
                                  const std::string& tool_instruction = {},
-                                 const json* unavailable_tools = nullptr) {
+                                 const json* unavailable_tools = nullptr,
+                                 const TemplateOpts* opts = nullptr) {
     std::string p;
     size_t start = 0;
     std::string sys;
@@ -516,8 +551,13 @@ inline std::string chatml_prompt(const std::vector<Msg>& msgs, const json& tools
     // thinking/effort state drives BOTH the reasoning_instructions line at the
     // system head and the generation prompt below (mirrors the template's
     // stateful ns_state). Template order: reasoning_instructions FIRST, then
-    // tools, then the client system content.
-    const ThinkToggles tt = scan_think_toggles(msgs, think);
+    // tools, then the client system content. Items 3/4: a request-driven base
+    // effort overrides the env/boot default, and auto_disable_thinking_with_tools
+    // starts thinking OFF when tools are present.
+    bool base_think = think;
+    if (opts && opts->auto_disable_thinking_with_tools && tools.is_array() && !tools.empty())
+        base_think = false;
+    const ThinkToggles tt = scan_think_toggles(msgs, base_think, opts ? opts->effort : -1);
     const std::string effort = tt.thinking ? reasoning_effort_line_level(tt.effort)
                                            : std::string();
     strip_think_markers(sys);
@@ -575,11 +615,11 @@ inline std::string chatml_prompt(const std::vector<Msg>& msgs, const json& tools
         }
         // normal (non-tool) message
         p += "<|im_start|>" + strip_ctrl(m.role) + "\n";
-        // assistant history reasoning block, jinja format:
-        //   <think>\n...\n</think>\n\n  then the plain content
-        if (m.role == "assistant" && !m.reasoning.empty()) {
-            // v22.3: if the client also sent the think block in text, strip it
-            // so the explicit reasoning is not double-rendered
+        // v22.3 E: EVERY assistant history turn is wrapped in <think>...</think>
+        // (even when reasoning is empty -> an empty think block). If the
+        // client also sent the think block in text, strip it so the block is
+        // not double-rendered.
+        if (m.role == "assistant") {
             strip_leading_think(content);
             p += "<think>\n" + strip_ctrl(m.reasoning) + "\n</think>\n\n";
         }
@@ -622,7 +662,9 @@ inline long request_max_chars(const json& body, const char* key) {
 
 inline std::string tool_call_text(const std::string& name, const json& args,
                                   long max_arg_chars = -1) {
-    std::string a = args.dump();
+    // v22.3 (item 7): a raw string argument is echoed verbatim (not JSON-encoded),
+    // matching the template's `_args = tc.arguments` when arguments is a string.
+    std::string a = args.is_string() ? args.get<std::string>() : args.dump();
     if (max_arg_chars < 0) { // no explicit request value -> env fallback
         const char* e = getenv("Q27_MAX_TOOL_ARG_CHARS");
         max_arg_chars = e ? atol(e) : 0;
@@ -634,6 +676,39 @@ inline std::string tool_call_text(const std::string& name, const json& args,
            "}\n</tool_call>";
 }
 
+// v22.3 (item 6): render an assistant tool call in the PROMPT using the active
+// tool dialect. For the XML dialect use the per-parameter form the template
+// emits (<function=NAME><parameter=K>v</parameter>...</function>); otherwise
+// fall back to the JSON tool_call_text.
+inline std::string tool_call_text_dialect(const std::string& name, const json& args,
+                                          long max_arg_chars = -1) {
+    if (max_arg_chars < 0) { // no explicit request value -> env fallback
+        const char* e = getenv("Q27_MAX_TOOL_ARG_CHARS");
+        max_arg_chars = e ? atol(e) : 0;
+    }
+    if (!tool_dialect_xml()) return tool_call_text(name, args, max_arg_chars);
+    auto trunc = [&](std::string v) -> std::string {
+        if (max_arg_chars > 0 && (long)v.size() > max_arg_chars)
+            return v.substr(0, (size_t)max_arg_chars) + "\n[TRUNCATED - original length " +
+                   std::to_string(v.size()) + " chars]";
+        return v;
+    };
+    std::string s = "<tool_call>\n<function=" + name + ">\n";
+    if (args.is_object()) {
+        for (auto it = args.begin(); it != args.end(); ++it)
+            s += "<parameter=" + it.key() + ">\n" +
+                 trunc(it.value().is_string() ? it.value().get<std::string>()
+                                              : it.value().dump()) +
+                 "\n</parameter>\n";
+    } else if (args.is_string()) {
+        s += trunc(args.get<std::string>()) + "\n";
+    } else {
+        s += trunc(args.dump()) + "\n";
+    }
+    s += "</function>\n</tool_call>";
+    return s;
+}
+
 inline std::string tool_response_text(const std::string& out,
                                       long max_response_chars = -1) {
     std::string o = out;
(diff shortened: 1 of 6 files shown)


signalnine:
thanks for this -- and particularly for the mapping table. having every behaviour
tied back to a documented row on the model card is what makes a port this size
reviewable at all, and it's not something i see often.

it merges clean onto current master and the full CPU suite is green here. two
things to settle first: one substantive, one where the description claims more
than the change does.

## the default effort

changing the default from `xhigh` to `medium` is the piece i can't take as-is.
it affects every request, and it quietly rebaselines every benchmark number we
have.

we measured this arm directly on 2026-08-19, three arms differing only in
`Q27_REASONING_EFFORT`, hidden tests, n=3 on the volatile tasks:

```
                              xhigh    medium     low
bench-task-queue              0.040     0.000   0.000
bench-time-tracker            0.173     0.147   0.067
bench-constraint-scheduler    0.965     0.667   0.982
bench-financial-ledger        1.000     1.000   0.775
MEAN hidden                   0.545     0.453   0.456
```

i called that a null at the time -- the per-trial outcomes are bimodal and n=3
samples a coin, so i don't claim xhigh is better. but it doesn't support
switching to medium either, and "zero injected tokens" is an efficiency argument
where the thing i'd want to protect is the score.

the template's default makes sense for the template. q27 isn't obliged to
inherit it, and if we change it i'd want a measured arm behind the decision
rather than upstream parity. could you split that into its own commit, or drop
it and leave `medium` reachable explicitly? the rest of the effort work (the
aliases, `none`/`off`, the inline `<|think_*|>` toggles) i have no argument
with.

## the P0 reasoning claim is narrower than stated

the PR says this "fixes the bug where a multi-turn agentic loop lost every prior
thought block after the first call". for the Anthropic path that isn't happening
today -- `api_common.h` already does:

```cpp
if (role == "assistant" && !think.empty())
    content = "<think>\n" + think + "\n</think>\n" + content;
```

so `thinking` blocks in history are rendered back into the prompt, and Claude
Code uses that path.

what IS missing is the OpenAI shape: `reasoning_content` as a sibling field on
an assistant message is never read on input, and your patch adds exactly that.
so the fix is real, just scoped to the OpenAI-compatible path rather than being
a general history-loss bug. worth correcting in the description, so nobody later
reads it as evidence the Anthropic path was broken.

## one question

the two-tier tool-error escalation injects system warnings based on heuristics
over a 120-char head. what did the false-positive rate look like in practice? a
call whose output legitimately contains `throw new` or a non-zero exit that
isn't a failure would get an unearned "try a fundamentally different approach"
in its prompt, and that's the kind of thing that's hard to notice and easy to
blame on the model. if you have session logs showing how often it fired, that'd
settle it.

happy to take the history rendering, the multi-system merge, the aliases and the
toggles once the default is split out. i haven't built the CUDA or Metal targets
against this yet -- i'll do both before merging, since `metal_server.cpp`
compiles `api_common.h` under `-Werror` and that has caught real breakage here
before.

signalnine:
built the two targets i said i would. CUDA is fine; **metal does not compile**.

`Msg` gains a third member (`std::string reasoning`), and `metal_server.cpp`
brace-initialises `Msg` at eight sites with two fields. that target builds with
`-Werror` and `-Wmissing-field-initializers` on apple clang, so every one of them
is a hard error:

```
src/metal/metal_server.cpp:182:60: error: missing field 'reasoning' initializer [-Werror,-Wmissing-field-initializers]
src/metal/metal_server.cpp:193:68: error: missing field 'reasoning' initializer
src/metal/metal_server.cpp:213:37: error: missing field 'reasoning' initializer
src/metal/metal_server.cpp:350:62: error: missing field 'reasoning' initializer
src/metal/metal_server.cpp:352:83: error: missing field 'reasoning' initializer
src/metal/metal_server.cpp:361:57: error: missing field 'reasoning' initializer
src/metal/metal_server.cpp:369:73: error: missing field 'reasoning' initializer
src/metal/metal_server.cpp:375:82: error: missing field 'reasoning' initializer
8 errors generated.
```

this is not on you to have caught -- you'd need an apple-silicon box, and the CPU
suites and the CUDA build are both blind to `src/metal/`, so `make test-tools`
is fully green on a tree where one of the two backends cannot build. that gap
has bitten this repo before and it is the reason i build metal on every parser
change now.

two ways to fix it, your call:

- give the member a default (`std::string reasoning{};` is enough for
  aggregate init to stop complaining), or
- add the third initialiser at those eight sites.

i'd take the default. it keeps `metal_server.cpp` out of a change that has
nothing to do with metal, and it means the NEXT member added to `Msg` doesn't
break that target either.

happy to run the build again once you push. the two review points from my earlier
comment still stand independently of this.

PierpaoloPernici:
Thanks for the thorough review! -- all three points are addressed in three new commits (`7058263`, `a10246b`, `e75b0ac`), now pushed, and the description is corrected :)

**The default effort** (`a10246b`)
Dropped. The `medium` fallback line is back to `xhigh`/`off` per boot dialect, so every existing client and every benchmark baseline is unchanged. Item 2 (`none/off` disables thinking end-to-end) is kept intact since that's the part you were fine with. `medium` remains reachable explicitly via `Q27_REASONING_EFFORT=medium`, the `reasoning_effort` field, or a `<|think_medium|>` tag. If we want to revisit the default later, I'd propose doing it as its own PR backed by a real measured arm -- the n=3 we have supports neither side, so it shouldn't ride along on a template port.

**Metal** (`7058263`)
Took your suggestion: `Msg::reasoning` now has an NSDMI (`std::string reasoning{};`) instead of touching the eight brace-init sites. That keeps `metal_server.cpp` out of this change and means the next member added to `Msg` doesn't break that target either. I don't have an Apple Silicon box, so I'd appreciate you running that build again -- but per standard clang behavior, a member with a default member initializer suppresses `-Wmissing-field-initializers` at aggregate init sites.

**The P0 description**
Corrected in the PR body: the fix is now scoped to the OpenAI path (`reasoning_content` was never read on input); you're right that the Anthropic path was already re-rendering `thinking` blocks, and the description no longer suggests otherwise.

**The tool-error escalation false-positive rate** (`e75b0ac`)
Honest answer: session logs are not persisted, so I have no fire-rate data to show. Instead, since `is_tool_response_failure` is pure, I probed it directly with realistic benign outputs -- including your exact worry vectors:

```
12 benign  -> 3 FALSE POSITIVES
   - exit code: 1 from grep no-match (UNIX convention, not a failure)
   - exit code: 1 from diff (differences found)
   - short build summary containing "error: none"
6 true failures -> 0 false negatives (traceback, command not found,
   fatal:, {"error": ...}, exit code: 2, "failed to" prefix)
```

So yes -- the instinct was right, and it's measurable: an otherwise-clean tool output can earn an unearned "try a fundamentally different approach". Two mitigations are included in this PR:

- `Q27_TOOL_ERROR_WARNINGS=0` flips the boot default off;
- per-request `tool_error_warnings: false` (top-level or `chat_template_kwargs`) overrides either way.

The default stays **ON** for template parity, but given your principle on item 1 was "no default change without a measured arm", I'm happy to flip this default too if you'd rather see escalation be opt-in -- say the word.

signalnine:
this is a good turnaround, thanks. built both targets against the new head:

**metal is green.** the NSDMI did it -- `q27-metal-server` links clean under
`-Werror` on apple clang (macOS 26.5, arm64) and the CPU suites pass there.
`q27-server` is complete under nvcc, and `make test-tools` is green here.
verified the default-effort revert too: line 304 is back to
`(tool_dialect_xml_default() ? "xhigh" : "off")`, so baselines are unchanged.

**on the false positives** -- you did the right thing there. i reproduced your
probe independently rather than take the numbers, with my own cases mixed in:

```
12 benign  -> 3 false positive(s)
   grep no-match (exit 1)
   diff found differences
   build summary 'error: none'
6 failures -> 0 false negative(s)
```

same three, same zero. finding them yourself and reporting them plainly is worth
more to me than a clean number would have been.

so: **saying the word.** please flip `tool_error_warnings` to default OFF.

the reasoning is the same one you applied to item 1, and you were right to apply
it there. this is a new default-on behaviour that injects text into the prompt,
and unlike the effort default we now have a measurement: 25% of benign outputs
trip it. `grep` returning 1 on no-match isn't an edge case either, it's one of
the most common tool results in an agentic loop -- so in practice an agent would
get told to "try a fundamentally different approach" after a routine search that
found nothing. that is a bad failure to debug, because the model looks like it
lost the plot and the prompt is where the plot went.

template parity is a fine reason to have the code. it isn't a reason to make it
the default when the measured cost lands on the hot path. the env var and the
per-request field mean anyone who wants template-identical behaviour still gets
it with one flag, and if a measured arm later shows the escalation helps, flipping
it back on is a one-line PR with evidence behind it.

everything else is verified and ready -- push that and i'll merge.

PierpaoloPernici:
Flipped -- `tool_error_warnings` now defaults to OFF (`8ec1db3`).

- `Q27_TOOL_ERROR_WARNINGS=1` re-enables at boot; the per-request `tool_error_warnings` boolean overrides either way.
- `chatml_prompt` falls back to the env default on the nullptr-opts path too, so no call site can accidentally keep the old behaviour.
- Tests rewritten around the new default: default-off, env opt-in, request-over-boot-default and request-over-env precedence.

Agreed on the framing: parity is a reason to have the code, not to default it onto the hot path when we have a measurement against it. If an arm ever shows the escalation helps, flipping back is a one-line PR with evidence behind it.

Pushed as `8ec1db3`; all CPU suites green here. Over to you for metal + merge + a beer! :)

signalnine:
merged. thanks -- this was a genuinely good PR to review, and a first one at
that.

verified on all three targets against the final head: `q27-metal-server` links
clean under `-Werror` on apple clang (macOS 26.5, arm64), CPU suites pass there
and here, `q27-server` complete under nvcc. checked the default flip directly
rather than from the diff:

```
tool_error_warnings default (env unset): OFF
grep no-match still classified a failure? yes
```

the second line being unchanged is the right outcome -- the detector wasn't the
problem, the default was, and you moved exactly the one thing that needed moving.
the nullptr-opts fallback was a good catch; i hadn't thought about the call sites
that pass no opts at all.

the part worth calling out is the false-positive probe. i asked for a fire rate,
you didn't have logs, and instead of saying so and moving on you exercised the
pure function and came back with "yes, it fires on benign output, here are the
three". that's the harder answer to give and it's the one that decided the
default. i reproduced it independently with extra cases and got the same 3 of 12
and the same zero false negatives.

all three review points landed as their own commits with the reasoning attached,
which made the merge commit easy to write honestly.

for context on where this lands: we've spent the last two days closing a
tool-call parity gap against llama.cpp on the same weights (now 1.0000 on both,
measured per-turn), and the remaining difference is task score -- q27 0.759 vs
llama.cpp 0.854. i tested the obvious suspect this morning, decode temperature,
and it went the wrong way: matching llama.cpp's temp 0.80 took q27 from 0.232 to
0.000 on the hardest task. currently testing quantization (q6 at 19.8 GB against
llama.cpp's Q5_K_M at 19.5 GB, a near-exact size match). your history-rendering
work plausibly touches that gap too, so if the quant arm comes back null i'll be
looking at prompt fidelity next, and this PR will be part of that picture.

beer noted :)

PR #40, opened 2026-08-30
server: /metrics Prometheus endpoint (opt-in via --enable-metrics)

PierpaoloPernici:
## What

A Prometheus text-exposition `GET /metrics` for the server, opt-in behind a
new `--enable-metrics` flag (default off), plus the telemetry plumbing it
needs: `src/metrics.h` (new), accumulation in the engine, and a `/v1/models`
context-length field. Full series reference and consumer recipes in
[docs/metrics-endpoint.md](docs/metrics-endpoint.md).

**Posture.** Without the flag the handler is never registered (`/metrics`
404s; `/health` untouched) and the per-request bookkeeping is skipped
entirely -- a deployment that does not want metrics keeps the zero-overhead
shape. With the flag the route is auth-exempt like `/health`: monitoring
pollers scrape without credentials, and an auth wall here would break every
unauthenticated scraper. The endpoint carries metadata-scale telemetry only
-- counts, latencies, occupancy, uptime -- never prompt text or generated
content (dispositioned in the SECURITY-MODEL addendum of 2026-08-30).

## The series

Per-API label `api=` (`chat` / `completions` / `messages` / `responses`):

| Series | Type | Meaning |
|---|---|---|
| `q27_requests_total` / `q27_requests_errors_total` | counter | Generation requests ([req] universe); `end=error` included |
| `q27_prompt_tokens_total` | counter | Prompt tokens sent to prefill (incl. cached) |
| `q27_prefill_computed_tokens_total` | counter | Prompt tokens actually computed |
| `q27_prefill_cached_tokens_total` | counter | Prompt tokens served from cache (RAM/ckpt/disk) |
| `q27_decode_tokens_total` | counter | Generated tokens delivered |
| `q27_queue_wait_seconds_total` | counter | Sum of slot-claim wait (µs resolution) |
| `q27_ttft_seconds` / `q27_e2e_seconds` / `q27_itl_seconds` | histogram | Prefill wall; arrival -> generation complete; per-request mean ITL |
| `q27_requests_inflight` / `q27_slots_total` | gauge | Slots claimed / configured |
| `q27_kv_usage_perc` / `q27_prefix_cache_entries` | gauge | Pool occupancy (0-1); cache entries (omitted when off) |
| `q27_spec_draft_tokens_total` / `q27_spec_accepted_tokens_total` / `q27_spec_accept_ratio` | counter/gauge | MTP accept-gate lanes and cumulative ratio |
| `q27_decode_tokens_processed_total` (+ prefill computed/cached) | counter | **Live** token counters, unlabeled |
| `q27_preemptions_total` | counter | Honest constant 0 -- q27 never preempts (FIFO admission) |
| `q27_uptime_seconds` | gauge | Server uptime |

One observation per generation request at the `[req]` point (`GenStats`
final there; `end=error` included), so `/metrics` counts match the `[req]`
log line-for-line. Latency definitions: **TTFT = `pf_ms`** (the prefill
wall), **E2E = arrival -> generation complete** (stamped at `[req]`, not
last byte), **ITL = `dec_ms / dec`** per request (observed only when
`dec > 0`).

## Live vs completion counters

The per-api totals stamp at request completion, which is a step function:
during a 2000-token decode every poll reads the same value, then +2000 at
the end. The `q27_*_tokens_processed_total` series exist for dashboards --
the engine bumps them at the single delivery point (`post_round` per
delivered token, `generate_prefill`'s success exit for prefill), so a poller
sees progress *during* generation:

Measured: a poller watching a 2048-token generation reads 79-115 t/s live
where the completion-based counter showed 0 then +2048. At rest the live
counters equal the sum of the labeled totals.

sparkDash-—-Multi-DGX-Spark-Monitoring-Dashboard

## MTP accept telemetry on both decode modes

`gate_lane_fired/acc` -> `q27_spec_accept_ratio` were greedy-only: the
sampled path (`spec_sample_round`, and the conductor's fused
`commit_outcome(sampled=true)`) deliberately updated nothing, so a
sampled-only serving profile (`--temp 1.0`, the measured recipe) read a
constant 0 despite speculating ~2.3 tokens/round. Both sampled paths now
mirror the greedy lane accounting, monitoring-only -- no depth-controller or
EMA feed (the sampled ceiling is fixed at 4; adaptive depth is greedy-only).
Sampled traffic measures 0.42-0.74 by request mix; 0 with traffic flowing
means the gate never fired.

`q27_preemptions_total` is an honest constant 0: admission is a FIFO queue
(a request that finds no free slot waits -- visible in
`q27_queue_wait_seconds_total`), a running request is never evicted.
Non-zero would mean a code path regressed into preemption; 0 is the healthy
steady state.

sparkDash-—-Multi-DGX-Spark-Monitoring-Q27

## Consistency and precision details

- **No torn scrapes.** `observe()` bumps separate atomics and a request can
  complete mid-render; without protection a scrape could show
  `+Inf != _count` and consumers drop the quantiles. Histograms are
  seqlocked (writer CAS-es into an odd seq -- three relaxed atomics; reader
  retries on a torn read), so the `+Inf == _count` invariant holds under
  traffic: 115/115 invariant-clean scrapes across three stress rounds.
- **Bucket tails sized for real work.** A 60k-token prefill takes ~25 s;
  TTFT edges now reach 60 s, E2E 300 s, ITL 100 ms, so p95 stays in finite
  buckets instead of the `+Inf` tail consumers report as "no data".
- **`%.6f` for scaled outputs.** Six significant digits round away
  microsecond deltas once totals grow, corrupting `rate()` on long-running
  servers.
- **Naming: `q27_*` only, no `vllm:`/`ds4_` compatibility aliases.** The
  semantics are q27's own (computed/cached prefill split, gate-lane MTP
  accounting, FIFO no-preemption) and versioned by this engine; borrowing
  another project's metric prefixes to reuse its dashboards couples this
  wire format to names it does not own.
- The MTP lane counters are plain per-engine longs (approximate-atomic
  reads, monotonic) -- documented and accepted for monitoring; everything
  else is atomic, gauges snapshot under `route_m`.

## Also in this branch

- `/v1/models` reports `max_model_len` (max slot ctx), the standard
  OpenAI-compatible context-length field.
- The Makefile server rules now depend on `src/metrics.h` (header-only edits
  previously produced stale binaries).

## Verification

- `make all` green (CLI, server, w8/w16 variants, test suites).
- Flag on/off verified on the live server: default off -> `/health` 200,
  `/metrics` 404; with the flag -> `/metrics` 200, invariant check 12
  series / 0 violations, generation unaffected (141-158 t/s, 2.3-2.9
  tokens/round).
- Percentiles finite on real traffic including two 24 s 60k-prompt prefills
  (TTFT p95 39.75 s interpolated in the 30-60 s bucket).

## Docs

- `docs/metrics-endpoint.md` -- the reference (posture, series table,
  observation semantics, consumer recipes).
- README (Serving), `docs/SECURITY.md` (Serving posture),
  `docs/SECURITY-MODEL.md` (Addendum 2026-08-30: the /metrics surface),
  `docs/BUILDLOG.md` (2026-08-30 (a)), `docs/BENCHMARKING.md` (the
  delta-accounting note).
**- As a downstream user I've already wired this endpoint into a sparkDash (https://github.com/MiaAI-Lab/sparkDash) q27 adapter — and if it lands, I plan to send MiaAI a PR for it too.**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,7 @@ build/test_manifest: tools/test_manifest.cu src/engine.cuh src/kv_pool.h src/pre
 	        build/pf4.o -o $@
 
 
-build/q27-server: src/server.cu src/engine.cuh src/kv_pool.h src/prefill_arena.h src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
+build/q27-server: src/server.cu src/engine.cuh src/metrics.h src/kv_pool.h src/prefill_arena.h src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
                   src/device_model.cu src/loader.cpp src/tokenizer.cpp src/api_common.h src/drift_capture.h src/stream_split.h src/markdown_lex.h \
                   src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/cuda_common.h src/toolgram.h \
                   src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h third_party/httplib.h build/pf4.o | build
@@ -293,7 +293,7 @@ build/i8g64_test: tools/i8g64_test.cu src/i8g64.cuh | build
 # zoo for every build, so the w8 delta is now small -- kept for the widest
 # fits (the historical role-set + 12x-zoo savings are engine-wide now).
 # Same sources, own binary.
-build/q27-server-w8: src/server.cu src/engine.cuh src/kv_pool.h src/prefill_arena.h src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
+build/q27-server-w8: src/server.cu src/engine.cuh src/metrics.h src/kv_pool.h src/prefill_arena.h src/conductor.h src/blocks.cu src/prefill.cu src/kernels.cu src/spec3.cu src/vgemm.cu \
                      src/device_model.cu src/loader.cpp src/tokenizer.cpp src/api_common.h src/drift_capture.h src/stream_split.h src/markdown_lex.h \
                      src/blocks.cuh src/kernels.cuh src/spec3.cuh src/prefill.cuh src/fdmma.cuh src/turbo3.cuh src/turbo5.cuh src/cuda_common.h src/toolgram.h \
                      src/depthctl.h src/toolconstrain.h src/tokenizer.h src/prefix_cache.h src/prefix_ram.h third_party/httplib.h build/pf4.o | build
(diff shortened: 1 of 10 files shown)


signalnine:
merged. reviewed the full diff plus a live smoke on the 5090 arm before merging: flag-off 404s with /health untouched, flag-on renders the full series, the +Inf == _count invariant held on every histogram scraped mid-generation, and the live token counters moved between polls during a decode exactly as the PR describes. extract_check byte-match, test-tools green, nvcc clean.

quality work -- the seqlock with the acquire-fence validation and the bounded-retry fallback is the right shape, the %.6f rationale is the kind of thing that usually gets learned the hard way later, and writing the SECURITY-MODEL disposition and the BUILDLOG entry in-voice saved me doing it.

two notes, neither blocking, logged so they're on record:

1. the engine's live_* atomics bump unconditionally -- the PR text says bookkeeping is "skipped entirely" without the flag, which is true of the [req]-point observe but not of the per-token relaxed fetch_adds in post_round. cost is a handful of uncontended host-side increments per round while the GPU does milliseconds of work, so it's noise, but if a future perf hunt ever stares at post_round, that's what those are.

2. the sampled-path gate-lane accounting means [req]'s gch/gnh/glf/gla now populate under sampled traffic where they read constant 0 before. that's a fix (and the /metrics accept ratio needs it), but anyone diffing [req] lines across versions under a sampled profile will see new fields moving -- noted here so the change is findable.

thanks for the contribution -- second one in, both landed clean.

